PR #95, opened 2022-10-13
Update to new type system

alexbaden:
Basic tests pass, but it would be good to run Modin tests too.


python/pyhdk/_execute.pxd
```diff
@@ -86,8 +86,8 @@ ctypedef shared_ptr[CResultSet] CResultSetPtr
 cdef extern from "omniscidb/QueryEngine/TargetMetaInfo.h":
   cdef cppclass CTargetMetaInfo "TargetMetaInfo":
     const string& get_resname()
-    const CSQLTypeInfo& get_type_info()
-    const CSQLTypeInfo& get_physical_type_info()
+    const CType& get_type_info()
+    const CType& get_physical_type_info()
```
ienkovich, line 90: The physical type was removed from the C++ structure.

python/pyhdk/_common.pyx
```diff
@@ -7,10 +7,10 @@ from libcpp.memory cimport unique_ptr, make_unique, shared_ptr
 from cython.operator cimport dereference
 
 cdef class SQLType:
```
ienkovich, line 9: It would be nice to also rename this one to `TypeId` similarly to `SQLTypeInfo`. I think we should avoid the impression that HDK is an SQL engine.

ienkovich (commented):
Got this error trying to run Modin tests:

```[2022-10-13 14:46:43.371979] [0x00007fc9b432a740] [info]    0 0 LogicalIR.cpp:370 Check failed: not_null || operand_lv->getType()->isIntegerTy(8)
Fatal Python error: Aborted
```

Main branch works OK in the same environment.

ienkovich:
@alexbaden The fix for Modin tests has been merged into the submodule. Could you please update the branch so I could make the final check?

alexbaden:
Done, and I renamed SQLType to TypeId as suggested.

ienkovich (approved):
Modin tests are OK no this branch.

PR #107, opened 2022-11-16
Fix the sanity test

leshikus:
Two self-hosted agents conflict with each other, and I have no access to one of them to add a label. So I moved the sanity tests to run on Github.


pbchekin (approved):
LGTM

PR #128, opened 2022-12-06
Make nvcc use conda gcc when possible

leshikus:
Subj

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,8 +24,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-        with:
-          submodules: true
+
+      - name: Free space
+        run: |
+          sudo rm -rf /usr/local/lib/android/sdk /opt/hostedtoolcache /usr/share/swift /opt/az
+          sudo apt-get purge -y gcc
 
       - name: Set env context
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -24,8 +24,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-        with:
-          submodules: true
+
+      - name: Free space
+        run: |
+          sudo rm -rf /usr/local/lib/android/sdk /opt/hostedtoolcache /usr/share/swift /opt/az
+          sudo apt-get purge -y gcc
```
Garra1980, line 31: non-related to Free space step?

leshikus: It's related, I remove gcc for conda build. As for the "rm" line, I will try to keep this step the same for all containers (and this removal is needed for docker containers).

leshikus: Now if someone again will add the call to OS gcc (not conda gcc), then the build will report that

leshikus: Should I also include this into PR https://github.com/intel-ai/hdk/commit/25adfaa10f466c3a94b1412d4de92f1e983a86f1 ?

.github/workflows/build.yml
```diff
@@ -70,12 +73,13 @@ jobs:
           CPU_COUNT: 2
           RUN_TESTS: 1
         run: |
+          sh -vx
```
Garra1980, line 76: debugging stuff?

.github/workflows/main.yml
```diff
@@ -62,3 +62,4 @@ jobs:
 
   sanity-test:
     uses: ./.github/workflows/build-docker-hdk.yml
+
```
Garra1980, line 65: let's remove this

PR #130, opened 2022-12-08
Support gcc12 and upgrade conda dev env

alexbaden:
Most changes are required to support gcc12 or silence warnings. The compiler has better static analysis and caught two places where null ptrs were accessed. Appears to be copy/paste issues and I made what seemed like reasonable fixes. 

@leshikus can we land this as is and then work on updating the Docker files, etc once we are sure everything works?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,7 +123,9 @@ set(CUDA_USE_STATIC_CUDA_RUNTIME ON CACHE STRING "Use static CUDA runtime")
 
 option(ENABLE_CUDA "Enable CUDA support" OFF)
 if(ENABLE_CUDA)
+  set(CMAKE_CUDA_ARCHITECTURES 60) # Set something here so that we can pick up nvcc
   enable_language(CUDA)
+  
   find_package(CUDAToolkit REQUIRED)
   # A temporary workaround for non-conda envs
   include_directories(${CUDAToolkit_INCLUDE_DIRS})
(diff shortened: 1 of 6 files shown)


omniscidb/IR/OpType.h
```diff
@@ -250,3 +248,5 @@ inline std::string toString(hdk::ir::WindowFunctionKind kind) {
 inline std::ostream& operator<<(std::ostream& os, hdk::ir::WindowFunctionKind kind) {
   return os << toString(kind);
 }
+
+}  // namespace hdk::ir
```
ienkovich, line 252: It was intentional to have `toString` and `operator<<` in the global namespace because we want to reload the global `toString` method and allow stream values without `using namespace hdk::ir`. We have similar methods for other enums in `hdk::ir::` namespace.

alexbaden: The problem w/ that seems to be in the logger `CHECK_EQ` macro, which is trying to access the `operator<<` overload and can't find it.

ienkovich: I'm surprised that moving it from the global namespace to `hdk::ir` helps to find it. Maybe we have some ambiguous overloads?

leshikus:
>> can we land this as is

I believe we better keep CI green. Technically you can temporarily comment failing checks in `main.yml`

leshikus (approved):
LGTM

alexbaden:
This has already been merged in other PRs.

PR #136, opened 2022-12-13
Move hdk::ir streaming operators to their namespace.

ienkovich:
Currently, stream operators for all `hdk::ir` enums are located near their `toString` overloads in the global namespace. But it creates problems with a lookup because it's different for regular functions and operators. This patch should fix the lookup issues by moving operators into the right place (enum's namespace).


alexbaden (approved):
Seems to have fixed the issues in the gcc12 branch. I rebased #130 over your commit, let's see how the tests go.

PR #139, opened 2022-12-15
DwarfBench integration

bagrorg:


omniscidb/scripts/mapd-deps-conda-dev-env.yml
```diff
@@ -49,4 +50,5 @@ dependencies:
   - binutils
   - llvm-spirv 14.0.*
   - folly 2022.11.07.00
+  - dpcpp_cpp_rt
```
alexbaden, line 53: Do we have to add this as a dependency for all environments? What does this do to our conda solve?

omniscidb/QueryEngine/CostModel/DataSources/DwarfBench.cpp
```diff
@@ -18,147 +18,76 @@
...
     const std::vector<AnalyticalTemplate>& templates) {
   Detail::DeviceMeasurements dm;
-  boost::filesystem::path dwarf_path = getDwarfBenchPath();
-
-  if (!boost::filesystem::exists(dwarf_path / "results")) {
-    boost::filesystem::create_directory(dwarf_path / "results");
-  }
-
   for (AnalyticalTemplate templ : templates) {
+    if (!isTemplateSupported(templ)) continue;    // Or should we throw an exception?
```
kurapov-peter, line 33: Since this will be a part of the calibration this should be an implementation detail for the query engine. You can throw an exception if we handle it in the calibration code, but I don't see much reason to. Trying to execute an unsupported template looks like incorrect application logic to me. How about a `CHECK` instead?

omniscidb/scripts/mapd-deps-conda-dev-env.yml
bagrorg: Since `DwarfBench` support have been made optimal in CMake, and dpc++ runtime is used only there, it would probably be more correct to take out `dpcpp_cpp_rt`. How could I do it more correctly?

omniscidb/scripts/mapd-deps-conda-dev-env.yml
```diff
@@ -11,6 +11,7 @@
 name: omnisci-dev
 channels:
   - conda-forge
+  - intel
```
kurapov-peter, line 14: I'd double-check this is the right channel to use for us.

omniscidb/scripts/mapd-deps-conda-dev-env.yml
kurapov-peter: @alexbaden, dwarf bench will require dpc++ runtime for all platforms. What happens to conda solve we need to find out :)

kurapov-peter (approved):
Trying to run a workflow.

kurapov-peter:
Tests pass for the PR, see CI status here https://github.com/intel-ai/hdk/tree/bagrorg/dbench

omniscidb/scripts/mapd-deps-conda-dev-env.yml
kurapov-peter: removed

kurapov-peter:
Reopened as #187

PR #140, opened 2022-12-16
Add ASAN test, first drop

leshikus:
The container still needs to be updated with new versions of LLVM, parquet, etc

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -8,6 +8,9 @@ on:
       name:
         type: string
         default: cpu
+      build-name:
+        type: string
+        default: build
       reset-cache:
         type: boolean
         default: false
@@ -28,7 +31,7 @@ jobs:
 
       - uses: actions/download-artifact@v3
         with:
-          name: ${{ env.RUN_STAMP }}-build
+          name: ${{ env.RUN_STAMP }}-${{ inputs.build-name }}
 
       - name: Unpack build files
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/test-docker.yml
```diff
@@ -47,8 +50,15 @@ jobs:
           docker exec hdk-build.${{ inputs.name }} chown -R ghrunner:ghrunner /_work/
           docker exec -u ghrunner hdk-build.${{ inputs.name }} tar -zxf /_work/build.tgz -C /_work/
 
-      - name: Test
-        run: docker exec -u ghrunner hdk-build.${{ inputs.name }} bash /_work/omniscidb/scripts/conda/test.sh
+      - name: Smoke test
+        if: inputs.name == 'cuda'
```
kurapov-peter, line 54: Is this a cuda-only flow?

.github/workflows/test.yml
```diff
@@ -68,6 +68,32 @@ jobs:
...
+
+        run: |
+          sudo sh -c 'echo 2 >/proc/sys/vm/overcommit_memory'
+          cd omniscidb
+
+          cd build/Tests # build cache
+          conda run -n ${{ env.CONDA_ENV }} ./ArrowBasedExecuteTest --build-rel-alg-cache=cache.txt
+
+          cd ../.. # rebuild with ASAN
+          conda run -n ${{ env.CONDA_ENV }} cmake -B ./build -S . -DENABLE_TESTS=on -DENABLE_CUDA=off -DENABLE_ASAN=on
```
kurapov-peter, line 84: Is this for saving up space or?

kurapov-peter (commented):
Looks fine, a couple questions.

.github/workflows/test-docker.yml
leshikus: this is a "self hosted" workflow; it applies now to two scenarios: CUDA tests and ASAN tests e g those we cannot run with github runners

leshikus: the strange way to pass "cuda" into a reusable workflow and query for it comes from two requirements/limitations:
1) Igor requests no cut&paste;
2) it was not possible (until recently) to call a function from another function

this results in the following approach

<common code>
<code for a cuda runner> if cuda
<code for an asan runner> if asan

.github/workflows/test.yml
leshikus: we need to run an unmodified test first to cache calcite queries and responses, this is done in this patch
then we need to recompile the whole project with ASAN and run the test with cached queries/responses, this is not fully finished

kurapov-peter: Ah, I see

kurapov-peter (approved):
LGTM

PR #142, opened 2022-12-16
Enable pull check

leshikus:
Subj

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -1,0 +1,82 @@
+name: Build
+on:
+  workflow_dispatch:
+    inputs:
+      reset-cache:
+        type: boolean
+        default: true
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


kurapov-peter (approved):

kurapov-peter:
Oh wait, it duplicates all the workflows.

kurapov-peter (changes requested):
There's something like https://github.com/marketplace/actions/skip-duplicate-actions for example to prevent the duplication. May be something else even better?

leshikus:
Thanks, will look into it

leshikus:
My previous experience shows that Igor generally does not approve use of third party actions. We may have smaller or different set of tests for pull requests.

kurapov-peter:
There is for example this method: https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/

leshikus:
This is a good text. AFAIU they reworked event object, so implementing this will require an additional investigation

PR #143, opened 2022-12-18
Upgrade to Calcite 1.32

alexbaden:
This PR has all the machinery to upgrade to Calcite 1.32. Major changes include:

* Remove the custom SqlToRelConverter in favor of a more targeted approach when setting up the Calcite planner. Previously, SqlToRel controlled the expansion of subqueries based on the properties of the SQL statement during conversion. Now, we control expansion based on either the SQL statement or the relational algebra composition, depending on the query. There was also some retry logic that was buried -- this logic has been pulled up into a top level try/catch (basically allows for "desugaring" patterns that can't be properly converted back to SQL after parsing). I did have to disable a few correlated subquery tests because I wasn't able to get them to work under the new scheme, but the queries there are not high priority. 
* Projected literals are (for the most part) now part of LogicalValues. This caused a few problems for our existing logical values handling, and exposed some bugs in the CAST framework. I was able to fix most issues, but we will need to address the issue of projecting BIGINT_MAX, which is our current EMPTY_KEY value. So, technically, we should restrict the range of BIGINT to be BIGINT_MIN + 1 and BIGINT_MAX - 1 (but we currently don't enforce the upper limit, at least not in logical values). 

I need to make some additional cleanup passes in the Java code -- it is pretty rough. But ArrowBasedExecuteTest and CorrelatedSubQueryTest are now passing locally for me with this set of patches. So, I am putting this up to run full CI with cleanup planned shortly.


alexbaden:
This has been cleaned up and is now ready for review.

PR #152, opened 2023-01-09
Simple toString() method for ArrowResultSet

akroviakov:
This commit adds a simple `toString()` method in [omniscidb/QueryEngine/ArrowResultSet.h](https://github.com/intel-ai/hdk/compare/main...akroviakov:hdk:akroviak/arrow_res_print?expand=1#diff-062062dc8d485c4c7f31d81129e4c641ee0c821b1170774e38c5fd16380488af) that utilizes Apache Arrow `RecordBatch` method `ToString()` for printing its contents.
It is tested in [omniscidb/Tests/ArrowBasedExecuteTest.cpp](https://github.com/intel-ai/hdk/compare/main...akroviakov:hdk:akroviak/arrow_res_print?expand=1#diff-d1c6de126cd483e52b9c42852207585c39b237e589c8b041f5a36f57fb37b69b), inside `c_arrow()` calls in the form of a simple print statement:
 `std::cout << arrow_omnisci_results->toString() << std::endl;`
 that is invoked by setting a flag parameter `print_res` to true in each `c_arrow()` call.

### DIFF
--- a/omniscidb/Tests/ArrowSQLRunner/ArrowSQLRunner.h
+++ b/omniscidb/Tests/ArrowSQLRunner/ArrowSQLRunner.h
@@ -100,6 +100,7 @@ void cta(const std::string& query_string, const ExecutorDeviceType device_type);
 
 void c_arrow(
     const std::string& query_string,
+    const bool print_res,
     const ExecutorDeviceType device_type,
     size_t min_result_size_for_bulk_dictionary_fetch =
         ArrowResultSetConverter::default_min_result_size_for_bulk_dictionary_fetch,
--- a/omniscidb/QueryEngine/ArrowResultSet.h
+++ b/omniscidb/QueryEngine/ArrowResultSet.h
@@ -139,6 +139,10 @@ class ArrowResultSet {
 
   const hdk::ir::Type* colType(size_t col_idx) const;
 
+  std::string toString() const{
+    return record_batch_->ToString();
+  }
+
   bool definitelyHasNoRows() const;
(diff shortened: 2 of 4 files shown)


omniscidb/Tests/ArrowSQLRunner/ArrowSQLRunner.h
```diff
@@ -100,6 +100,7 @@ void cta(const std::string& query_string, const ExecutorDeviceType device_type);
 
 void c_arrow(
     const std::string& query_string,
+    const bool print_res,
```
kurapov-peter, line 103: I'd avoid changing this interface. If you do want to do that, please consider making it an optional parameter with a default `false` value.

kurapov-peter (changes requested):
Please make sure your code is properly formatted (clang-format).

omniscidb/Tests/ArrowSQLRunner/ArrowSQLRunner.h
ienkovich: I agree that adding a new arg for debug prints is not the best option (even if it is optional). We might want to use debug prints in many places and such args would spread all over the code. Also, having debug prints that can only be enabled through source code modifications is not really convenient. Consider using debug logging or debug config in this case. If you are not sure what would be the best option here, then simply adding a new debug function is also fine for the PR, you don't have to utilize it.

kurapov-peter:
Moved to #160

PR #160, opened 2023-01-13
Added toString methods for ResultSet and ArrowResultSet

akroviakov:
This PR adds "toString" methods for `ResultSet` and `ArrowResultSet`.
`ArrowResultSet::toString()` calls Arrow's `RecordBatch::ToString()`.
`ResultSet::contentToString()` walks over rows and prints columns of the current row.
___________________
Changes outside of `ResultSet` and `ArrowResultSet` files:
`ExtractFromTime` now has methods to retrieve string representation of a timestamp in different formats.
___________________
Clarifying question | possible bug(?):
`ExtractFromTime` contains functions to extract certain time units from a timestamp. When the extraction unit is less than a second (milli, micro, nano), the extraction basically looks like this (e.g., for millisecond):
`unsigned_mod(lcltime, kSecsPerMin * kMilliSecsPerSec)`
so we basically take the remainder of a division by units per _minute_ (for millisecond it is 60000). In `ResultSet` timestamps are stored as seconds with milli/micro/nano scale meaning timestamp extension by 3/6/9 digits respectively. 
The question is why does the subsecond scale extraction requires division by units-per-minute and not by units-per-second? 
These functions for subsecond scales are already used in `ExtractFromTime(hdk::ir::DateExtractField field, const int64_t timeval)`, so if there is no issue with these functions, what is the expected format of `lcltime` for subsecond scales (other functions for year, month, day, hour, minute, second work fine as is)?

### DIFF
--- a/omniscidb/QueryEngine/ExtractFromTime.cpp
+++ b/omniscidb/QueryEngine/ExtractFromTime.cpp
@@ -21,9 +21,9 @@
 
 #include "ExtractFromTime.h"
 
-#include <sstream>
-#include <iomanip>
 #include <cmath>
+#include <iomanip>
+#include <sstream>
 
 #ifndef __CUDACC__
 #include <cstdlib>  // abort()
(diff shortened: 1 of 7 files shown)


omniscidb/QueryEngine/ExtractFromTime.cpp
```diff
@@ -312,3 +316,66 @@ DEVICE int64_t ExtractFromTime(hdk::ir::DateExtractField field, const int64_t ti
...
+    case hdk::ir::TimeUnit::kMicro:
+      return getStrTimeStampSecondsScaled(tstamp, kMicroSecsPerSec);
+    case hdk::ir::TimeUnit::kNano:
+      return getStrTimeStampSecondsScaled(tstamp, kNanoSecsPerSec);
+    case hdk::ir::TimeUnit::kMonth:
+      return getStrMonthFromSeconds(tstamp);
+    case hdk::ir::TimeUnit::kDay:
+      return getStrDayFromSeconds(tstamp);
+  }
+}
```
kurapov-peter, line 381: I think you need to wrap this code with the ```#ifndef __CUDACC__```. The code is compiled with nvcc and causes failures upon linking (see CI).

A missing EOL here btw.

kurapov-peter (changes requested):

kurapov-peter (approved):
LGTM. The cuda test failure seems to be unrelated.

PR #161, opened 2023-01-13
Directly reference Join nodes in ColumnRef.

ienkovich:
Currently, when a column of a `Join` node is referenced, appropriate `ColumnRef` actually references one of its input nodes, and `Join` nodes are not used in `ColumnRef` expressions. That was probably convenient in some analysis code and visitors, but for HDK IR as a public API, it creates very nonobvious inconsistency. During my previous refactorings, I was preparing a switch to the direct usage of `Join` nodes in `ColumnRef` expressions. This is the final step that actually starts doing it.

I intentionally didn't modify `LeftDeepInnerJoin` case, because it's not used right now by default and this node is going to be removed anyway.

### DIFF
--- a/omniscidb/IR/Node.cpp
+++ b/omniscidb/IR/Node.cpp
@@ -611,11 +611,12 @@ ExprPtrVector getNodeColumnRefs(const Node* node) {
                 LogicalUnion,
                 LogicalValues,
                 Filter,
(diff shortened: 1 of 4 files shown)


kurapov-peter (approved):
LGTM

PR #162, opened 2023-01-13
Introduce QueryBuilder

ienkovich:
This PR introduces a public API for building Relational Algebra DAG. The aim of this API is to provide an access to low-level HDK IR API but in a much more friendly way comparing to direct nodes and expressions creation via their constructors.

Many builder methods have a lot of overloads to allow a Python-like variety for input operand types. I'm not sure we need that many overloads, so we might revise it in the future and reduce it.

The coverage is not even close to 100%. For nodes it doesn't cover TableFunction and LogicalUnion. For expressions it actually doesn't cover a lot of things including string expressions, subqueries, window functions, function operator, etc. But it is still good enough to cover NYC Taxi and TPCH Q1 and Q3.

This API will be exposed in PyHDK to provide similar interfaces for Python.

The patch is quite big, but most of it is TPCH data generator (I had to add it instead of data itself because of license issues) and builder tests.


ienkovich:
@vlad-penkin Could you please check that TPCH data generator is added correctly from the licenses point of view?

ienkovich:
@leshikus I've tried to reproduce the ASAN failure but actually failed to build with enabled ASAN in conda environment with some errors in folly (and I see the same errors trying ASAN build on the main branch). Is ASAN build enabled for docker only for now? Do we use folly in ASAN build?

leshikus:
>> Is ASAN build enabled for docker only for now?

I've mostly used conda version for ASAN and never tried it in docker

kurapov-peter (commented):
I've skimmed through most of the PR except for the builder itself. Looks reasonable as far as I can tell. Just a couple comments.

omniscidb/QueryEngine/ArrowResultSetConverter.cpp
```diff
@@ -1281,7 +1281,9 @@ std::shared_ptr<arrow::DataType> get_arrow_type(const hdk::ir::Type* type,
     case hdk::ir::Type::kText:
       return arrow::utf8();
     case hdk::ir::Type::kDecimal:
-      return arrow::decimal(type->as<hdk::ir::DecimalType>()->precision(),
+      // No reason to use 256-bit decimals since we always import 64-bit values.
+      CHECK_EQ(type->size(), 8);
+      return arrow::decimal(std::min(type->as<hdk::ir::DecimalType>()->precision(), 38),
```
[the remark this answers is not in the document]
ienkovich: 38 is the max precision for a 64-bit decimal. In our result, we can get higher precision especially when multiplication is used.

omniscidb/SchemaMgr/SchemaMgr.cpp
```diff
@@ -0,0 +1,82 @@
...
+  mgr_by_schema_id_.emplace(schema_id, schema_provider);
+}
+
+const SchemaProvider* SchemaMgr::getMgr(int db_id) const {
+  auto it = mgr_by_schema_id_.find(getSchemaId(db_id));
+  if (it != mgr_by_schema_id_.end()) {
+    return it->second.get();
+  }
+  return nullptr;
+}
```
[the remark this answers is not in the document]
ienkovich: Fixed

omniscidb/IR/Context.cpp
```diff
@@ -311,6 +313,138 @@ class ContextImpl {
...
+  const Type* typeFromString(const std::string& val) {
+    auto val_lower = boost::trim_copy(boost::to_lower_copy(val));
+    std::smatch match_res;
+
+    // Null type.
+    if (val_lower == "nullt") {
+      return null();
+    }
+    // Boolean.
+    if (std::regex_match(val_lower, match_res, std::regex("bool(\\[nn\\])?"))) {
```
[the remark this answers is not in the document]
ienkovich: Not really. The best documentation, for now, is the appropriate test suite and docstrings in the proposed python API: https://github.com/intel-ai/hdk/pull/170/files#diff-d8b8cca193af0d6768b0e0809b22ef0871563d378ab087d5f6a95e0863019eafR1198

omniscidb/QueryBuilder/QueryBuilder.h
```diff
@@ -0,0 +1,656 @@
...
+class BuilderExpr {
+ public:
+  BuilderExpr(const BuilderExpr& other) = default;
+  BuilderExpr(BuilderExpr&& other) = default;
+
+  BuilderExpr& operator=(const BuilderExpr& other) = default;
+  BuilderExpr& operator=(BuilderExpr&& other) = default;
+
+  BuilderExpr name(const std::string& name) const;
+  const std::string& name() const { return name_; }
```
kurapov-peter, line 35: These two:
```
  BuilderExpr name(const std::string& name) const;
  const std::string& name() const { return name_; }
```
look like a getter/setter pair at a first glance, might benefit from renaming.

omniscidb/QueryBuilder/QueryBuilder.h
```diff
@@ -0,0 +1,656 @@
...
+  static SortDirection parseSortDirection(const std::string& val);
+  static NullSortedPosition parseNullPosition(const std::string& val);
+
+ protected:
+  std::variant<int, std::string, BuilderExpr> field_;
+  SortDirection dir_;
+  NullSortedPosition null_pos_;
+};
+
+class BuilderNode {
```
kurapov-peter, line 222: Could you please elaborate on the abstractions' relation to each other and the reasoning behind them?

omniscidb/QueryBuilder/QueryBuilder.h
```diff
@@ -0,0 +1,656 @@
...
+
+namespace hdk::ir {
+
+class InvalidQueryError : public Error {
+ public:
+  InvalidQueryError() {}
+  InvalidQueryError(std::string desc) : Error(std::move(desc)) {}
+};
+
+class QueryBuilder;
```
kurapov-peter, line 22: Does a `QueryBuilder` belong in `hdk::ir` namespace and not just `hdk`?

omniscidb/QueryBuilder/QueryBuilder.h
```diff
@@ -0,0 +1,656 @@
...
+class BuilderNode;
+class ExprRewriter;
```
kurapov-peter, line 24: What is the rewriter exposed for?

omniscidb/QueryBuilder/QueryBuilder.h
```diff
@@ -0,0 +1,656 @@
...
+  BuilderExpr ge(const std::string& val) const;
+
+  BuilderExpr at(const BuilderExpr& idx) const;
+  BuilderExpr at(int idx) const;
+  BuilderExpr at(int64_t idx) const;
+
+  BuilderExpr rewrite(ExprRewriter& rewriter) const;
+
+  BuilderExpr operator!() const;
+  BuilderExpr operator-() const;
```
kurapov-peter, line 159: What does it semantically mean "to subtract two builder expressions"?

omniscidb/QueryBuilder/QueryBuilder.h
```diff
@@ -0,0 +1,656 @@
...
+  BuilderExpr time(const std::string& val) const;
+  BuilderExpr timestamp(const std::string& val) const;
+
+ protected:
+  friend class BuilderExpr;
+  friend class BuilderNode;
+
+  BuilderNode scan(TableInfoPtr table_info) const;
+
+  Context& ctx_;
```
kurapov-peter, line 530: Is it possible to reuse a chain of operators built with a different builder? Assuming they share the schema and data source.

kurapov-peter (commented):
Added some inline questions/comments on the API. A general question: what do we provide expression comparison operators for?

omniscidb/QueryBuilder/QueryBuilder.h
ienkovich: How about `BuilderExpr rename(const std::string& name) const`? E. g. `node.ref("value").sum().rename("total_value")`.

omniscidb/QueryBuilder/QueryBuilder.h
ienkovich: `BuilderNode` wraps `hdk::ir::Node` and provides an additional building interface to it. `BuilderExpr` makes the same for `hdk::ir::Expr`.

omniscidb/QueryBuilder/QueryBuilder.h
ienkovich: `QueryBuilder` interfaces are purely to manipulate IR, so I guess this is the right namespace. I'm not sure about these two-level namespaces in the first place, but apparently, we already have class names intersections (e.g. we have two `ColumnRef` classes)

omniscidb/QueryBuilder/QueryBuilder.h
ienkovich: For `BuilderExpr::rewrite` interface

omniscidb/QueryBuilder/QueryBuilder.h
ienkovich: This is a unary minus. It allows to use it as `neg_val = -node.ref("value")`. There is also a binary operator which allows you to build subtractions like `diff = node["x"] - node["y"]`.

omniscidb/QueryBuilder/QueryBuilder.h
ienkovich: IR contexts were introduced for the isolation of IRs built in different threads. If you work on the same thread and have two builders working with the same context, then you can mix their expressions, but I don't see why you would need it.

So far, we don't actually manage contexts nicely in our code and always use the same default context. That means we shouldn't try to build multiple queries from different threads. I hope it will change some days and we start creating query-scoped contexts. For Python, I think we will continue using a single global context for all queries.

omniscidb/QueryBuilder/QueryBuilder.h
kurapov-peter: Yeah, I was thinking of `rename` or `named`. I like `rename`.

ienkovich: Done

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+                              << expr_->type()->toString();
+  }
+  auto agg =
+      makeExpr<AggExpr>(builder_->ctx_.fp64(), AggType::kAvg, expr_, false, nullptr);
+  auto name = name_.empty() ? "avg" : name_ + "_avg";
+  return {builder_, agg, name, true};
+}
+
+BuilderExpr BuilderExpr::min() const {
+  if (!expr_->type()->isNumber() && !expr_->type()->isDateTime()) {
```
kurapov-peter, line 507: No strings?

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+  if (res_type->isInteger() && res_type->size() < 8) {
+    res_type = builder_->ctx_.int64(res_type->nullable());
+  }
+  auto agg = makeExpr<AggExpr>(res_type, AggType::kSum, expr_, false, nullptr);
+  auto name = name_.empty() ? "sum" : name_ + "_sum";
+  return {builder_, agg, name, true};
+}
+
+BuilderExpr BuilderExpr::count(bool is_distinct) const {
+  if (!expr_->is<hdk::ir::ColumnRef>()) {
```
kurapov-peter, line 541: Would a scan over the whole table be represented by a column ref? I'm thinking of something similar to `select (*)`.

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+      {"single value", AggType::kSingleValue}};
+  static const std::unordered_set<std::string> distinct_names = {
+      "count_dist", "count_distinct", "count dist", "count distinct"};
+  auto agg_str_lower = boost::algorithm::to_lower_copy(agg_str);
+  if (!agg_names.count(agg_str_lower)) {
+    throw InvalidQueryError() << "Unknown aggregate name: " << agg_str;
+  }
+
+  auto kind = agg_names.at(agg_str_lower);
+  if (kind == AggType::kApproxQuantile && val == HUGE_VAL) {
```
kurapov-peter, line 635: What value is huge? 3? :)

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+  return mul(builder_->cst(val, builder_->ctx_.fp32(false)));
+}
+
+BuilderExpr BuilderExpr::mul(double val) const {
+  return mul(builder_->cst(val, builder_->ctx_.fp64(false)));
+}
+
+BuilderExpr BuilderExpr::div(const BuilderExpr& rhs) const {
+  try {
+    auto bin_oper = Analyzer::normalizeOperExpr(
```
kurapov-peter, line 1073: Would it check for a constant div by 0?

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+  }
+  auto expr_ptrs = collectExprs(exprs);
+  return proj(expr_ptrs, fields);
+}
+
+BuilderNode BuilderNode::proj() const {
+  std::vector<int> col_indexes(node_->size());
+  for (int i = 0; i < (int)col_indexes.size(); ++i) {
+    col_indexes[i] = i;
+  }
```
kurapov-peter, line 1606: iota?

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+                              size_t limit,
+                              size_t offset) const {
+  return sort(std::vector<BuilderSortField>({field}), limit, offset);
+}
+
+BuilderNode BuilderNode::sort(const std::vector<BuilderSortField>& fields,
+                              size_t limit,
+                              size_t offset) const {
+  std::vector<SortField> collation;
+  for (auto field : fields) {
```
kurapov-peter, line 2168: May cause a deep copy. Is this intentional?

kurapov-peter (commented):
Added some comments on the `QueryBuilder` implementation.

omniscidb/QueryBuilder/QueryBuilder.h
kurapov-peter: What is the `rewrite` interface used for? I think I saw it for input replacement, but what are the major use cases?

kurapov-peter:
One more general question. Do we intend to allow any IR extensions?

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+  return nullptr;
+}
+
+ExprPtr getRefByName(const Project* proj, const std::string& col_name) {
+  for (size_t i = 0; i < proj->size(); ++i) {
+    if (proj->getFieldName(i) == col_name) {
+      return getNodeColumnRef(proj, (unsigned)i);
+    }
+  }
+  return nullptr;
```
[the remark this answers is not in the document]
ienkovich: This is handled on a higher level. This code is caused by the lack of field names in join nodes. For joins, we use the same method to search for a field in its input nodes. If we get `nullptr` from all inputs then an exception is thrown.

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -0,0 +1,3072 @@
...
+        throw InvalidQueryError() << "Empty field names are not allowed";
+      }
+      auto pr = names.insert(expr.name());
+      if (!pr.second) {
+        throw InvalidQueryError() << "Duplicated field name: " << expr.name();
+      }
+    }
+  }
+
+  // Build the resulting vector adding suffixes to auto-names when needed.
```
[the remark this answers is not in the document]
ienkovich: I'd say, that for non-empty names suffix is `_N`, and for empty names, the suffix is `expr_N`, we never add prefixes to non-empty names.

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: We don't support min/max aggregate for strings.

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: No, column ref is just a reference to a column in a scalar expression. E.g. `SELECT x FROM table;` would have a single column reference in a projection. `SELECT * FROM table;` would require as many column references, as the table has columns.

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: `HUGE_VAL` is a constant used as `+INF` double value. In this function, we use this value by default to allow users to skip `val` arg for most of the aggregates. And if `APPROX QUANTILE` aggregate is used but `val` is missing, then an exception is thrown.

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: No. Normalization only makes sure we produce an expression we can handle in our codegen. It is mostly to add required casts to operands and choose the proper type for the result. I don't know if it makes sense to check for a division by zero because just having it is not an error (expression might be conditional and never execute).

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: STL is so rich ) Will use it

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: It's a bug. Thanks. Will fix.

ienkovich:
> One more general question. Do we intend to allow any IR extensions?

I don't see any reason to have IR extensions for now. We hardly can provide convenient interfaces to integrate user-defined expressions into our codegen and it's even harder to do that for nodes. UDF is currently the only way to add user-defined features.

kurapov-peter (approved):
LGTM. I think I went through most of it, but we probably need yet another pair of eyes here.

PR #163, opened 2023-01-13
Refactor const ref args in Node classes ctors and setters.

ienkovich:
During work on `Node` classes, I replaced many constant reference arguments with by-value arguments. This is because constructors and setters are going to store them in the object anyway, so using by-value arguments allows a caller to avoid a copy if he decides to move-create or emplace-create arguments (he has no such option with constant reference arguments).

Some constant reference cases were missed and move-construction is not always used. This minor refactoring fixes that.


kurapov-peter (approved):
LGTM!

PR #166, opened 2023-01-18
[README] Update readme build info.

Devjiu:
This commit updates build instructions and removes obsolete information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,35 +18,35 @@ We are committed to supporting a baseline set of functionality on all x86 CPUs, 
 
 `ArrowStorage` is currently the default (and only available) HDK storage layer. `ArrowStorage` provides storage support for [Apache Arrow](https://github.com/apache/arrow) format data. The storage layer must be explicitly initialized:
 
-```
+```python
 import pyhdk
 storage = pyhdk.storage.ArrowStorage(1)
 ```
 
-The parameter applied to the `ArrowStorage` constructor is the database ID. The database ID allows storage instances to be kept logically separate.
+The parameter applied to the `ArrowStorage` constructor is the database ID. The database ID allows storage instances to *be* kept logically separate.
 
 `ArrowStorage` automatically converts Arrow format datatypes to `omniscidb` datatypes. Some variable length types are not yet supported, but scalar types are available. `pyarrow` can be used to convert Pandas DataFrames to Arrow:
 
-```
-       at = pyarrow.Table.from_pandas(
-            pandas.DataFrame({"a": [1, 2, 3], "b": [10, 20, 30]})
-        )
+```python
+at = pyarrow.Table.from_pandas(
+    pandas.DataFrame({"a": [1, 2, 3], "b": [10, 20, 30]})
+)
 ```
 
 The arrow table can then be imported using the Arrow storage interface.
 
-```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,44 +72,47 @@ The complete flow is as follows:
...
 
-```
-mkdir build
-cd build
-cmake -DENABLE_CONDA=on ..
-make -j
-make install
+```bash
+    mkdir build && cd build
+    cmake -DENABLE_CUDA=off ..
```
ienkovich, line 103: `Heterogeneous` is a part of our library name but build instructions are for the CPU-only version. Should we propose CUDA enabled build as the default option and describe CPU-only version as an alternative for cases when the CUDA toolkit is not available?

README.md
```diff
@@ -72,44 +72,47 @@ The complete flow is as follows:
...
-     conda activate hdk-dev
-```
+If you meet issues during build refer to [`.github/workflows/build.yml`](https://github.com/intel-ai/hdk/blob/b0bc9ca3f599f648bcc1a8e1eb49c0f02e1f7aca/.github/workflows/build.yml)
 
 ## Test
 
 Python tests can be run from the python source directory using `pytest`. 
+
+```bash
+     pytest python/tests/
```
ienkovich, line 117: You mentioned it fails due to the missing Modin dependency. If we are fixing the documentation, then it would be nice to include all the required steps to successfully run tests.

Devjiu: Fixed

Devjiu:
Looks like tests are not stable - Build / sanity-test / build (push) Failing after 50m. 

Also, I would suggest skipping a full rebuild for such minor changes.

README.md
```diff
@@ -59,57 +59,119 @@ Three high level components are required to execute a query:
...
+```bash
+mkdir build && cd build
+cmake -DENABLE_CUDA=on  ..
+make -j 
+make install
 ```
 
+### Issues
+
+If you meet issues during build refer to [`.github/workflows/build.yml`](https://github.com/intel-ai/hdk/blob/b0bc9ca3f599f648bcc1a8e1eb49c0f02e1f7aca/.github/workflows/build.yml). This file describes compilation steps used for CI build.
```
kurapov-peter, line 133: I think a relative link to the file would be more stable here so that you don't need to update it.

kurapov-peter (approved):
Looks fine, just one comment

README.md
Devjiu: Thanks, fixed.

README.md
```diff
@@ -59,57 +59,119 @@ Three high level components are required to execute a query:
...
+## Pytest logging 
+
+To enable logging enable with this function: 
+```python
+pyhdk.initLogger(debug_logs=True)
+```
+
+In `setup_class(..)` body.
+
+Logs are by default located at `mapd_log/` folder.
```
ienkovich, line 177: @kurapov-peter @alexbaden @leshikus Do we have any scripts working with logs or depending on the log dir in some other way? It would be nice to finally rename this dir.

ienkovich (approved):

README.md
kurapov-peter: I don't think we do except for CI log uploading maybe?

Devjiu: updated according to #180

PR #171, opened 2023-01-20
Add test to detect improper working dir.

ienkovich:
With this test, we enforce pytest run from the root HDK directory. This is to state that we don't support pytest run from arbitrary places to easily locate resource files (like test CSV files) in python tests.

### DIFF
--- a/python/tests/test_work_dir.py
+++ b/python/tests/test_work_dir.py
@@ -1,0 +1,13 @@
+#
+# Copyright 2023 Intel Corporation.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+
+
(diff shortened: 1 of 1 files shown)


kurapov-peter:
Why force it at all?

ienkovich:
> Why force it at all?

In C++ we use macroses to specify test files' location. For Python, we cannot do a similar thing through cmake. Working dir for tests depends on where tests were run from and it makes it hard to find data files (there is already a PR working with the taxi data sample). Fixing pytest run dir is the simplest solution. This test is to provide a user with a better description of the error rather than stating that some files are missing.

We are open to alternate solutions.

PR #172, opened 2023-01-23
L0 parallel code generation and execution for Taxi queries.

kurapov-peter:
This PR enables simple group-by-agg queries multithreaded execution for the L0 path. It allows for executing all NYC Taxi queries to run on the GPU. No shared memory is used currently. Group initialization happens on the CPU.

### DIFF
--- a/omniscidb/L0Mgr/L0Mgr.cpp
+++ b/omniscidb/L0Mgr/L0Mgr.cpp
@@ -146,6 +146,9 @@ L0Device::L0Device(const L0Driver& driver, ze_device_handle_t device)
                                                 ZE_COMMAND_QUEUE_PRIORITY_NORMAL};
   L0_SAFE_CALL(
       zeCommandQueueCreate(driver_.ctx(), device_, &command_queue_desc, &queue_handle));
+  L0_SAFE_CALL(zeDeviceGetProperties(device_, &props_));
+  CHECK_EQ(ZE_DEVICE_TYPE_GPU, props_.type);
+  L0_SAFE_CALL(zeDeviceGetComputeProperties(device_, &compute_props_));
 
   command_queue_ = std::make_shared<L0CommandQueue>(queue_handle);
 }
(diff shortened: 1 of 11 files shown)


omniscidb/L0Mgr/L0Mgr.cpp
```diff
@@ -174,6 +177,13 @@ std::unique_ptr<L0CommandList> L0Device::create_command_list() const {
   return std::make_unique<L0CommandList>(res);
 }
 
+unsigned L0Device::maxGroupCount() const {
```
alexbaden, line 180: Can we use an explicitly sized type here?

omniscidb/QueryEngine/Compiler/genx.ll
```diff
@@ -50,3 +49,29 @@ define i64 @agg_count(i64* %agg, i64 %val) {
...
     ret i64 %old
 }
+
+define i64 @agg_count_skip_val(i64* %agg, i64 noundef %val, i64 noundef %skip_val) {
+    %no_skip = icmp ne i64 %val, %skip_val
+    br i1 %no_skip, label %.noskip, label %.skip
+.noskip:
+;;  todo: %old = call i64 @agg_count(i64* %agg, i64 %val)
+    %ld = load i64, i64* %agg
+    %old = atomicrmw add i64* %agg, i64 1 monotonic
```
alexbaden, line 59: do we need this atomic?

omniscidb/QueryEngine/Descriptors/QueryMemoryDescriptor.cpp
```diff
@@ -1059,8 +1059,13 @@ bool QueryMemoryDescriptor::blocksShareMemory() const {
 }
 
 bool QueryMemoryDescriptor::lazyInitGroups(const ExecutorDeviceType device_type) const {
+#ifdef HAVE_L0
+  return false; /* L0 path currently does not provide lazy group by buffers initialization
```
alexbaden, line 1063: Won't this disable lazy init groups for CUDA too? Or are we not planning to build CUDA + L0?

kurapov-peter: There's no way of building CUDA + L0 currently, so CUDA is not affected.
>Or are we not planning to build CUDA + L0

We do, and there're some changes that minimize the gap we have now, but I have no immediate plans for enabling both in a single build right now. Over time it became much easier to enable, so we might want to do that soon.

omniscidb/QueryEngine/Compiler/genx.ll
kurapov-peter: Not really, a regular `add` works just fine here. At least for non-grouped aggregations that do not use shared memory ;). I plan to revisit it once we have shared memory turned on (which I'm working on now).

omniscidb/L0Mgr/L0Mgr.cpp
kurapov-peter: Done

alexbaden (approved):

PR #173, opened 2023-01-23
[CSV parsing] Empty columns support.

Devjiu:
This commit support empty column from csv parsing.


alexbaden:
What happens if the schema is explicitly not null, e.g.:

```
create table t (x int not null, y int, z int);
```
and then import your test CSV file.

ienkovich:
> What happens if the schema is explicitly not null, e.g.:
> 
> ```
> create table t (x int not null, y int, z int);
> ```
> 
> and then import your test CSV file.

I'm not sure we really check for NULL values when importing into a non-nullable column. Worth testing.

Devjiu:
> What happens if the schema is explicitly not null, e.g.:
> 
> ```
> create table t (x int not null, y int, z int);
> ```
> 
> and then import your test CSV file.

Added check, that should cover this case, also test "ImportCsv_KnownSchema_NullsColumn_NullableTypeHeader" describes similar scenario.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -470,14 +470,19 @@ void ArrowStorage::appendArrowTable(std::shared_ptr<arrow::Table> at, int table_
   for (auto& frag : fragments) {
     frag.metadata.resize(at->columns().size());
   }
-
```
ienkovich, line 473: Please try to avoid unrelated changes. Do you intentionally remove this vertical space? If yes, then what for?

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -470,14 +470,19 @@ void ArrowStorage::appendArrowTable(std::shared_ptr<arrow::Table> at, int table_
...
   mapd_shared_lock<mapd_shared_mutex> dict_lock(dict_mutex_);
   threading::parallel_for(
       threading::blocked_range(0, (int)at->columns().size()), [&](auto range) {
         for (auto col_idx = range.begin(); col_idx != range.end(); col_idx++) {
           auto col_info = getColumnInfo(db_id_, table_id, col_idx + 1);
           auto col_type = col_info->type;
           auto col_arr = at->column(col_idx);
+          LOG(INFO) << "append Arrow: " << col_type->toString()
```
ienkovich, line 480: INFO is a default logging level. I don't think this message should go to the log by default. Also, it lacks log context because we don't have much logging in the ArrowStorage, and therefore is not very useful. If it was only for your personal debugging, then remove it.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -852,10 +856,12 @@ void ArrowStorage::compareSchemas(std::shared_ptr<arrow::Schema> lhs,
     auto lhs_type = lhs_fields[i]->type();
     auto rhs_type = rhs_fields[i]->type();
 
-    if (!lhs_type->Equals(rhs_type)) {
-      throw std::runtime_error("Mismatched type for column "s + lhs_fields[i]->name() +
-                               ": "s + lhs_type->ToString() + " vs. "s +
-                               rhs_type->ToString());
+    if (!lhs_type->Equals(rhs_type) && !(lhs_type->id() == arrow::Type::NA) &&
```
ienkovich, line 859: Why `!(a == b)` instead of `a != b`? Also, the NA type is not allowed for LHS.

omniscidb/ArrowStorage/ArrowStorageUtils.cpp
```diff
@@ -1082,6 +1089,9 @@ std::shared_ptr<arrow::DataType> getArrowImportType(hdk::ir::Context& ctx,
         return list(getArrowImportType(ctx, elem_type));
       }
     }
+    case hdk::ir::Type::kNull:
+      throw std::runtime_error(
+          "Null should be converted to ArrowType according to getTargetImportType(...).");
```
ienkovich, line 1094: Runtime errors are for users and should explain to users what happened. This is not a useful error message, the previously used message about unsupported type is more informative.

omniscidb/DataMgr/Encoder.cpp
```diff
@@ -60,7 +60,7 @@ Encoder* Encoder::Create(Data_Namespace::AbstractBuffer* buffer,
         case 8:
           return new NoneEncoder<int64_t>(buffer);
         default:
-          CHECK(false);
+          CHECK("Unhandled primitive type size" && false);
```
ienkovich, line 63: I don't think we need any messages here. But anyway, it is done through streaming for our checks: `CHECK(false) << "Error message."`. And if still want to have a message here, then better to print the type (`toString` gives the full info about the type) causing the problem.

omniscidb/ArrowStorage/ArrowStorageUtils.cpp
```diff
@@ -196,6 +199,8 @@ std::shared_ptr<arrow::ChunkedArray> replaceNullValuesImpl(
   std::shared_ptr<arrow::Array> array;
   if constexpr (std::is_same_v<T, bool>) {
     array = std::make_shared<arrow::Int8Array>(arr->length(), std::move(resultBuf));
+  } else if (arr->type()->id() == arrow::Type::NA) {
```
ienkovich, line 202: I have a doubt here. What if we have an existing table with int or some other column and try to import NULL column into it? It's supposed to work, but here we always create doubles. I'm not sure if it later matters and this type is used anywhere, but a better way here would be to make arrow type out of the known C type. 
```
using ResultArrowType = typename arrow::CTypeTraits<T>::ArrowType;
using ArrayType = typename arrow::TypeTraits<ResultArrowType>::ArrayType;
```

omniscidb/Tests/ArrowStorageTest.cpp
```diff
@@ -829,6 +829,90 @@ TEST_F(ArrowStorageTest, ImportCsv_UnknownSchema_Numbers_NoHeader) {
...
+  table_options.fragment_size = 32'000'000;
+
+  std::string exp_message =
+      "Null values used in non-nullable type: "s + ctx.int32(false)->toString();
+
+  try {
+    ArrowStorage::CsvParseOptions parse_options;
+    parse_options.header = false;
+    storage.appendCsvData("1, 10.0", tinfo->table_id, parse_options);
+    storage.appendCsvFile(getFilePath("nulls_header.csv"), "table1");
```
ienkovich, line 907: You can simply use `ASSERT_THROW` as in other examples.

ienkovich (changes requested):
Overall the patch looks good. I have several minor comments about unrelated changes, styling, etc.

omniscidb/ArrowStorage/ArrowStorage.cpp
Devjiu: ok, I will remove this change.

omniscidb/ArrowStorage/ArrowStorage.cpp
Devjiu: Ok, I will remove it, but I think there should be more debug logging. Arrow storage types and issues during append and types conversion not logged.

omniscidb/ArrowStorage/ArrowStorage.cpp
Devjiu: LHS not allowed to be NA, due to usage of `compareSchemas`. I think `compareSchemas` can be utility/static and should be symmetric to lhs/rhs.

omniscidb/ArrowStorage/ArrowStorageUtils.cpp
Devjiu: Ok, changed to ArrayType for consistency and have Null -> fp64 conversion at single case. Testscase with appendance of Null column to schema discovered from previous csv with int. 
 I'm not sure if there will be any changes in logic after this fix.

omniscidb/Tests/ArrowStorageTest.cpp
Devjiu: Added extra-macro. `ASSERT_THROW` not able to check message of an error.

ienkovich (approved):
PR looks good to go! You might want to rebase it first, CI should be green if rebased on the current main branch.

Devjiu:
> PR looks good to go! You might want to rebase it first, CI should be green if rebased on the current main branch.

Hm, looks like there was Null on non-nullable type. 
```
[  FAILED  ] Select.GroupEmptyBlank
C++ exception with description "Null values used in non-nullable type: DICT32(TEXT[NN])[16777272]" thrown in the test body
```

ienkovich:
> Hm, looks like there was Null on non-nullable type.

That's probably a sign to leave the null-check part for another PR. I don't think that in this test we actually have imported NULL value. The test is for an empty string and the reference value has no NULLs. Also, changing the type to allow NULLS in the test table causes the test to fail. So, in this case, NULL is probably interpreted as an empty string on import/execution/conversion.

Devjiu:
@alexbaden PTAL your notice (verification for nulls depending on schema) not covered fully. String conversion to null looks a bit difficult to support in current PR. Should we crate an issue with this and remove this WA in further PR? Or is current WA fine?

alexbaden (approved):
Let's create an issue and handle later. The empty handling here looks good.

PR #180, opened 2023-01-31
Use hdk_log as default log dir.

ienkovich:
This doesn't allow configuring dir for log4j (or disabling Calcite logs) but it is still a slight improvement compared to the current `${sys:MAPD_LOG_DIR}` which is more annoying. This also puts all logs into a single place (at least by default).


alexbaden (approved):

kurapov-peter (approved):
Thanks!

PR #188, opened 2023-02-09
Minor extensions to QueryBuilder.

ienkovich:
There are a few changes to QueryBuilder that I had to introduce during Python API implementation.
- Filter and sort over scan do not expose virtual columns anymore. The idea is that the virtual column should go into the result only when it's explicitly used by the user
- Default ctors were added for some structures. It makes their usage in Cython much easier because in Cython declaration with initialization is always split into a declaration + assignment
- Ceil and floor functions were added (floor is required to support Python __floordiv__ '//' operator)

### DIFF
--- a/omniscidb/QueryBuilder/QueryBuilder.cpp
+++ b/omniscidb/QueryBuilder/QueryBuilder.cpp
@@ -262,6 +262,14 @@ class InputNodesCollector
   }
 };
 
+class InputColIndexesCollector
+    : public ExprCollector<std::unordered_set<int>, InputColIndexesCollector> {
+ protected:
+  void visitColumnRef(const hdk::ir::ColumnRef* col_ref) override {
+    result_.insert(col_ref->index());
+  }
+};
+
 void expandAllowedInput(const Node* node,
                         std::unordered_set<const Node*>& allowed_nodes) {
   allowed_nodes.insert(node);
@@ -482,6 +490,8 @@ JoinType parseJoinType(const std::string& join_type) {
 }
 
 }  // namespace
+
+BuilderExpr::BuilderExpr() : builder_(nullptr) {}
 
 BuilderExpr::BuilderExpr(const QueryBuilder* builder,
                          ExprPtr expr,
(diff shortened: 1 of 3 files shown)


omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -1621,9 +1662,19 @@ BuilderNode BuilderNode::proj(const ExprPtrVector& exprs,
...
-BuilderNode BuilderNode::filter(BuilderExpr condition) const {
+BuilderNode BuilderNode::filter(const BuilderExpr& condition) const {
   checkExprInput(condition, {node_.get()}, "filter");
-  auto filter = std::make_shared<Filter>(condition.expr(), node_);
+  // Filter-out virtual column if it is not used on the filter.,
+  auto base = node_;
+  if (node_->is<Scan>()) {
+    auto used_cols = InputColIndexesCollector::collect(condition.expr());
+    int cols_to_proj =
+        used_cols.count((int)(node_->size() - 1)) ? node_->size() : node_->size() - 1;
```
kurapov-peter, line 1672: Can `(int)(node_->size() - 1)` be equal to `-1`?

omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -2187,7 +2238,18 @@ BuilderNode BuilderNode::sort(const std::vector<BuilderSortField>& fields,
   // Sort over scan is not supported.
   auto base = node_;
   if (node_->is<Scan>()) {
-    base = proj().node();
+    // Filter out rowid column if it's not used in the sort.
+    bool uses_rowid = false;
+    for (auto& col : collation) {
```
kurapov-peter, line 2243: Does this loop look like a `find_if`?

kurapov-peter (approved):
Looks good, just some minor notes. Thanks for making PRs shorter! :)

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: It is a scan, so it cannot be empty and its size is at least 2.

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: It appears that `find_if` option looks more verbose and probably less readable:
```
   bool uses_rowid =
        std::find_if(collation.begin(),
                     collation.end(),
                     [rowid_idx = node_->size() - 1](const& SortField field) {
                       return col.getField() == rowid_idx;
                     }) != collation.end();
```

PR #191, opened 2023-02-14
Move to default gcc version (in a docker)

leshikus:
Subj


leshikus:
It seems the default version for ubuntu 20 was in fact gcc-9, the fix is invalid

PR #192, opened 2023-02-14
Implement PyHDK QueryExpr, QueryNode and HDK API except data import.

ienkovich:


python/pyhdk/_builder.pyx
```diff
@@ -0,0 +1,722 @@
...
+    return expr, sort_dir, null_pos
+
+  @property
+  def is_scan(self):
+    return self.c_node.node().get().isNode[CScan]()
+
+  @property
+  def size(self):
+    res = self.c_node.size()
+    # For scans we hide virtual column
```
kurapov-peter, line 560: I didn't follow closely, `c_node.size()` isn't this a call to C++ API? And if so, doesn't it already contain the logic for subtracting `1` in case of a scan?

ienkovich: In C++ we actually do not hide the virtual column in such a manner. So we do all these modifications for virtual columns in Python.

kurapov-peter: Do you think this should be the user's responsibility? I'd think it's a rather specific feature for modin, isn't it?

ienkovich: Well, `rowid` virtual column is useful and is used by Modin. I don't want users to 'see' this column when they ask for node size or use an integer index to choose a column. I wrote some examples and it feels really confusing when this virtual column affects size and indexing (e.g. `ht[-1]` would give you the last column in most cases but for scans, it would be the `rowid` column). I believe the current Python implementation is the right way to go (we still can reference `rowid` column by its name). Probably, we should adjust C++ API implementation to have a match.

kurapov-peter: Yeah, I think consistent semantics is important. Anyway, it's not related to this PR. LGTM!

kurapov-peter (approved):

PR #193, opened 2023-02-15
Add spirv translator discovery to hdk complete build

kurapov-peter:
We'll need the translator for llvm versions < 15.


leshikus (approved):
LGTM

PR #195, opened 2023-02-17
Intel gpu support: aggregations

kurapov-peter:
This adds a separate enabling test for ease of development and some preparational changes for shared memory utilization.

### DIFF
--- a/omniscidb/L0Mgr/L0MgrNoL0.cpp
+++ b/omniscidb/L0Mgr/L0MgrNoL0.cpp
@@ -115,30 +115,37 @@ void L0Manager::synchronizeDevices() const {
 
 size_t L0Manager::getMaxAllocationSize(const int device_num) const {
   CHECK(false);
+  return 0u;
 }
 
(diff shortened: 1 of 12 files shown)


omniscidb/L0Mgr/L0MgrNoL0.cpp
```diff
@@ -133,6 +133,14 @@ unsigned L0Manager::getMinEUNumForAllDevices() const {
   CHECK(false);
 }
 
+bool L0Manager::hasSharedMemoryAtomicsSupport() const {
+  CHECK(false);
```
ienkovich, line 137: Could you please add `return` statements in this and similar implementations? I know this `CHECK` leads to `abort` call and the return statement would be a dead code but compilers complain about missing returns and on Windows, it is treated as an error, not just a warning.

kurapov-peter: Right, thanks for reminding me!

lmontigny (approved):
LGTM

PR #197, opened 2023-02-20
Add l0 tests

leshikus:
Subj

if `runs-on: [self-hosted, linux, $var]` the solution may be much simpler; yet it cannot

### DIFF
--- a/.github/workflows/test-l0-docker.yml
+++ b/.github/workflows/test-l0-docker.yml
@@ -1,0 +1,67 @@
+name: Docker test
+on:
+  workflow_call:
+    inputs:
+      cuda_compiler_version:
+        type: string
+        default: None
+      name:
+        type: string
+        default: cpu
+      build-name:
+        type: string
+        default: build
+      reset-cache:
+        type: boolean
+        default: false
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test-l0-docker.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Start the container
+        run:  |
+          docker run -id --name hdk-build.${{ inputs.name }} --network host --device /dev/dri:/dev/dri hdk-build.${{ inputs.name }}:latest
+          docker exec hdk-build.${{ inputs.name }} mkdir -p /_work/
+          docker cp build.tgz hdk-build.${{ inputs.name }}:/_work/
+          docker exec hdk-build.${{ inputs.name }} chown -R ghrunner:ghrunner /_work/
+          docker exec -u ghrunner hdk-build.${{ inputs.name }} tar -zxf /_work/build.tgz -C /_work/
+
+      - name: Smoke test
+        if: inputs.name == 'cuda'
```
kurapov-peter, line 54: This doesn't seem to run any tests.

Issue #183, opened 2023-02-02
ArrowResultSetConverter ignores LIMIT and OFFSET


ienkovich:
ResultSet has two fields related to LIMIT and OFFSET request features:
```
  size_t drop_first_;
  size_t keep_first_;
```
Looks like we completely ignore them (or at least when we use columnar conversion) in ArrowResultSetConverter. As a result, we have more output rows in the resulting Arrow table than expected. Can be reproduced by executing SQL query with LIMIT and OFFSET and converting the result to an Arrow table.

PR #199, opened 2023-02-20
[ArrowResultSet] Support LIMIT and OFFSET

Devjiu:
This commit takes into account the limit and offset parameters in the ArrowResultSet.
Also added several tests to check limit/offset with ascending/descending
ordering and joins to ResultSetArrowConversion suit.

Resolves: https://github.com/intel-ai/hdk/issues/183

### DIFF
--- a/omniscidb/QueryEngine/ArrowResultSetConverter.cpp
+++ b/omniscidb/QueryEngine/ArrowResultSetConverter.cpp
@@ -342,7 +342,7 @@ void convert_column(ResultSetPtr result,
         reinterpret_cast<const uint8_t*>(chunk_ptr), buf_size, result));
   }
 
-  CHECK_EQ(total_row_count, entry_count);
+  CHECK_LE(total_row_count, entry_count);
 
   std::vector<std::shared_ptr<arrow::Array>> fragments(values.size(), nullptr);
 
@@ -786,18 +786,30 @@ size_t convert_rowwise(
     std::vector<std::shared_ptr<std::vector<bool>>>& null_bitmap_seg,
     const std::vector<bool>& non_lazy_cols,
     const size_t start_entry,
-    const size_t end_entry) {
+    const size_t end_entry,
+    bool is_truncated = false) {
   const auto col_count = results->colCount();
   CHECK_EQ(value_seg.size(), col_count);
   CHECK_EQ(null_bitmap_seg.size(), col_count);
   const auto local_entry_count = end_entry - start_entry;
   size_t seg_row_count = 0;
+  size_t limit = results->getLimit();
+  size_t offset = results->getOffset();
   for (size_t i = start_entry; i < end_entry; ++i) {
+    if (is_truncated && seg_row_count >= offset + limit) {
+      break;
+    }
+
     auto row = results->getRowAtNoTranslations(i, non_lazy_cols);
     if (row.empty()) {
       continue;
     }
     ++seg_row_count;
+
+    if (is_truncated && seg_row_count <= offset) {
+      continue;
+    }
+
     for (size_t j = 0; j < col_count; ++j) {
       if (!non_lazy_cols.empty() && non_lazy_cols[j]) {
         continue;
@@ -1204,7 +1216,21 @@ std::shared_ptr<arrow::Table> ArrowResultSetConverter::getArrowTable(
 
     std::atomic<size_t> row_count = 0;
 
-    {
+    if (results_->isTruncated()) {
+      // TODO(dmitriim) This approach can be optimized for large offsets.
+      // We'll process it without tbb for now, because it's hard to keep track of the
+      // number of rows received across all threads.
+      auto timer = DEBUG_TIMER("fetch data single thread with limit/offset");
+      row_count += convert_rowwise(results_,
+                                   builders,
+                                   device_type_,
+                                   column_value_segs[0],
+                                   null_bitmap_segs[0],
+                                   columnar_conversion_flags,
+                                   0,
+                                   entry_count,
+                                   results_->isTruncated());
+    } else {
       auto timer = DEBUG_TIMER("fetch data in parallel_for");
       threading::parallel_for(
           static_cast<size_t>(0), entry_count, stride, [&](size_t start_entry) {
--- a/omniscidb/QueryEngine/ResultSet.cpp
+++ b/omniscidb/QueryEngine/ResultSet.cpp
@@ -797,6 +797,10 @@ size_t ResultSet::getLimit() const {
   return keep_first_;
 }
 
+size_t ResultSet::getOffset() const {
+  return drop_first_;
+}
+
 const std::vector<std::string> ResultSet::getStringDictionaryPayloadCopy(
     const int dict_id) const {
   const auto sdp = row_set_mem_owner_->getOrAddStringDictProxy(dict_id);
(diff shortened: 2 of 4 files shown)


omniscidb/Tests/ResultSetArrowConversion.cpp
```diff
@@ -385,6 +385,96 @@ TEST(ArrowTable, EmptySelection) {
   ASSERT_EQ(table->num_rows(), (int64_t)0);
 }
 
+TEST(ArrowTable, LimitedSelection) {
+  auto res =
+      runSqlQuery("select bi from test limit 4 offset 1;", ExecutorDeviceType::CPU, true);
```
ienkovich, line 390: This and similar queries might not produce the desired `ResultSet`. This is because for such projections it's simple to not include skipped rows in the execution phase. So, you should use sort and/or groupby to produce different ResultSet layouts.

ienkovich (changes requested):

omniscidb/QueryEngine/ArrowResultSetConverter.cpp
```diff
@@ -375,7 +375,10 @@ void convert_column(ResultSetPtr result,
                          : std::make_shared<NumArray>(chunk_rows_count, values[idx]);
   });  // threading::parallel_for
 
-  out = std::make_shared<arrow::ChunkedArray>(std::move(fragments));
+  const size_t first_entry = result->rowIterator(false, false).getCurrentRowBufferIndex();
```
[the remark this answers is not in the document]
Devjiu: Fixed check, please.

omniscidb/Tests/ResultSetArrowConversion.cpp
Devjiu: Fixed, check please.

omniscidb/QueryEngine/ResultSet.cpp
```diff
@@ -1498,9 +1502,34 @@ std::vector<std::pair<const int8_t*, size_t>> ResultSet::getChunkedColumnarBuffe
...
   retval.reserve(1 + appended_storage_.size());
 
-  retval.emplace_back(
-      storage_->getUnderlyingBuffer() + storage_->getColOffInBytes(column_idx),
-      storage_->binSearchRowCount());
+  auto rows_to_fetch = storage_->binSearchRowCount();
+  if (isTruncated()) {
+    rows_to_fetch = std::min(rows_to_fetch, keep_first_);
+  }
+  retval.emplace_back(storage_->getUnderlyingBuffer() +
```
ienkovich, line 1509: This code is incorrect for cases when `drop_first_ + keep_first_ > storage_->binSearchRowCount()`. It would reference past the storage buffer. Big offsets might force you to skip one or more storages but your code doesn't do it.

omniscidb/Tests/ResultSetArrowConversion.cpp
```diff
@@ -385,6 +385,111 @@ TEST(ArrowTable, EmptySelection) {
...
+
+  ScopeGuard reset = [prev_enable_columnar_output, prev_enable_lazy_fetch] {
+    config().rs.enable_columnar_output = prev_enable_columnar_output;
+    config().rs.enable_lazy_fetch = prev_enable_lazy_fetch;
+  };
+
+  config().rs.enable_columnar_output = true;
+  config().rs.enable_lazy_fetch = false;
+  auto res = runSqlQuery(
+      "SELECT * FROM test INNER JOIN join_table ON test.i=join_table.i offset 4 limit 2;",
```
ienkovich, line 443: For all non-groupby tests use `test_chunked` or run tests twice, once on `test` and another one on `test_chunked`. The chunked table holds the same data but is split into different fragments. This will produce ResultSets with multiple storages and improve your testing coverage.

omniscidb/QueryEngine/ArrowResultSetConverter.cpp
```diff
@@ -786,18 +786,33 @@ size_t convert_rowwise(
...
     auto row = results->getRowAtNoTranslations(i, non_lazy_cols);
     if (row.empty()) {
       continue;
     }
     ++seg_row_count;
+
+    if (is_truncated && seg_row_count <= offset) {
+      continue;
+    }
+    if (is_truncated && seg_row_count > offset + limit) {
```
ienkovich, line 812: We don't need to look for the next valid entry if converted enough rows. So please move this exit condition to the very beginning of the loop.

ienkovich (changes requested):
This patch uses the right idea and should work for ResultSets with single storage. But I think it doesn't handle multiple storages correctly for columnar conversion. Also, tests don't include multiple storages case. Please, work in this direction.

omniscidb/QueryEngine/ArrowResultSetConverter.cpp
Devjiu: Agree,

omniscidb/QueryEngine/ResultSet.cpp
Devjiu: I think it will be incorrect in case of `drop_first_ >= storage_->binSearchRowCount()`, if only sum is larger I am expecting data form offset to the end of table. (also I am checking it in `OrderedLimitOverSizeSelection` test)

Devjiu: The logic here is getting a little hard to understand. Maybe it's better to wrap in some method, not sure what's the best approach here.

ienkovich: `storage_` doesn't hold the whole table, so it's legal to drop more rows than `storage_` has.

Devjiu: Should be failed with an error or empty selection?

Devjiu: BTW, why it's illegal? We simply should skip first storage and use next one.

ienkovich: I didn't say it was illegal, I said the opposite. I mentioned missing storage skip in my first comment.

omniscidb/QueryEngine/ResultSet.cpp
```diff
@@ -1498,17 +1504,65 @@ std::vector<std::pair<const int8_t*, size_t>> ResultSet::getChunkedColumnarBuffe
...
-  for (auto& chunk_uptr : appended_storage_) {
+  for (auto& storage_uptr : appended_storage_) {
     const int8_t* ptr =
-        chunk_uptr->getUnderlyingBuffer() + chunk_uptr->getColOffInBytes(column_idx);
-    size_t row_count = chunk_uptr->binSearchRowCount();
-    retval.emplace_back(ptr, row_count);
+        storage_uptr->getUnderlyingBuffer() + storage_uptr->getColOffInBytes(column_idx);
+    size_t row_count = storage_uptr->binSearchRowCount();
+    curr_storage_rows_to_fetch = row_count;
+    if (curr_storage_end_row == 0)
```
ienkovich, line 1539: `curr_storage_end_row == 0` means we have to stop, right? If we got all the required rows from the first storage, then `curr_storage_end_row` should be 0 here and we should stop.

Overall, it's hard to read the code because of misleading variable names. `curr_storage_rows_to_fetch` doesn't hold the number of rows to fetch. `curr_storage_start_row` and `curr_storage_end_row` sound like iterators to something but in fact they are not. `expected_row_count` is named and used as if it holds a total number of rows to fetch, but it holds the total number of rows in the result set instead and `(total_handled_rows >= expected_row_count)` check is useless.

Could you please just use `rows_to_fetch` and `rows_to_skip` variables instead of the 5 variables you currently have? Initialize them like
```
size_t rows_to_skip = getOffset();
size_t rows_to_fetch = getLimit() ? std::min(getLimit(), rowCount() - getOffset()) : rowCount() - getOffset();
```
Then use them to compute a number of rows to skip/fetch from each storage and adjust them accordingly, stopping when `rows_to_fetch` is zero. I believe that would make the code more readable.

Also, please pay attention to your tests. Try to trigger current errors in tests before fixing them.

omniscidb/Tests/ResultSetArrowConversion.cpp
```diff
@@ -385,6 +387,231 @@ TEST(ArrowTable, EmptySelection) {
...
+    config().rs.enable_columnar_output = prev_enable_columnar_output;
+    config().rs.enable_lazy_fetch = prev_enable_lazy_fetch;
+  };
+
+  config().rs.enable_columnar_output = true;
+  config().rs.enable_lazy_fetch = false;
+  auto res = runSqlQuery(
+      "SELECT * FROM test_chunked offset 3 limit 7;", ExecutorDeviceType::CPU, true);
+
+  ASSERT_EQ(res.getRows()->rowCount(), (int64_t)3);
```
ienkovich, line 513: I definitely see problems in the current columnar conversion implementation. That makes me think these tests don't actually give you the required result set layouts. Please add asserts for `isChunkedZeroCopyColumnarConversionPossible()`, `getLimit()`, and `getOffset()` to make sure you have what you expect.

ienkovich (changes requested):

omniscidb/QueryEngine/ResultSet.cpp
```diff
@@ -1498,17 +1504,65 @@ std::vector<std::pair<const int8_t*, size_t>> ResultSet::getChunkedColumnarBuffe
...
+  size_t expected_row_count = rowCount();
+  // LOG(ERROR) << "storage";
+  // LOG(ERROR) << "raw row fetch: " << curr_storage_rows_to_fetch
+  //            << " start: " << curr_storage_start_row << " end: " <<
+  //            curr_storage_end_row;
+  if (curr_storage_rows_to_fetch <= curr_storage_start_row) {
+    curr_storage_start_row -= curr_storage_rows_to_fetch;
+  } else {
+    size_t to_fetch = std::min(curr_storage_rows_to_fetch - curr_storage_start_row,
+                               curr_storage_end_row - curr_storage_start_row);
```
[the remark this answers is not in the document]
Devjiu: Agree, error can be triggered only when limit/offset in first storage. If it's in next one everything ok.

omniscidb/Tests/ResultSetArrowConversion.cpp
Devjiu: donedone

omniscidb/QueryEngine/ResultSet.cpp
Devjiu: done

ienkovich (approved):
Looks good! Thanks!

omniscidb/Tests/ResultSetArrowConversion.cpp
```diff
@@ -145,7 +145,7 @@ void compare_columns(const std::array<TYPE, len>& expected,
   const arrow::ArrayVector& chunks = actual->chunks();
 
   TYPE null_val = null_builder<TYPE>();
-
+  auto total_row_count = 0;
```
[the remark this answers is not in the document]
Devjiu: agree, done

kurapov-peter (approved):
LGTM

PR #203, opened 2023-02-22
Add support for hash join on rowid

### LINKED ISSUE #165: Unable to join tables if there are more than 1000 rows
Code to reproduce:
```
import numpy as np
import modin.pandas as pd
import modin.config as cfg

cfg.IsExperimental.put(True)
cfg.Engine.put("Native")
cfg.StorageFormat.put("Hdk")

data = np.random.randint(0, 100, size=(1001, 99))
df = pd.DataFrame(data)
df["new"] = pd.Series(list(range(len(df))))
print(df)
```

Error message:
```
modin/experimental/core/execution/native/implementations/hdk_on_native/hdk_worker.py:76: in _executeRelAlgJson
    res = rel_alg_executor.execute()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   RuntimeError: Hash join failed, reason(s): Could not build a 1-to-1 correspondence for columns involved in equijoin | Cannot join on rowid | Cannot fall back to loop join for non-trivial inner table size

_sql.pyx:103: RuntimeError
```

alexbaden:
Support join on rowid by skipping the hash table build and computing the row offset in the probe. Offset computation consists of null handling and range handling. For an empty table, we use roughly the expression range of (0, -1) which always returns -1 inside the newly added `rowid_hash_join_idx` runtime function. 

There is one limitation -- because of this approach of not building the table we cannot support a query of the form `t1.rowid = t2.rowid AND .... `. Loop joins still work of course. If this query becomes something we need to support with hash join, we can push the secondary join expression into filters and run a mini-loop join on each row that passes the rowid "hash". 

Added a bunch of tests but could probably use some more coverage. Open to ideas. The way I force the hash table builder to skip the build is by marking it empty during construction. This is kind of a hack, but I think it's ok for now with the checks I've added. 

Closes #165

### DIFF
--- a/omniscidb/QueryEngine/JoinHashTable/PerfectJoinHashTable.cpp
+++ b/omniscidb/QueryEngine/JoinHashTable/PerfectJoinHashTable.cpp
@@ -434,7 +434,19 @@ ColumnsForDevice PerfectJoinHashTable::fetchColumnsForDevice(
   for (const auto& inner_outer_pair : inner_outer_pairs_) {
     const auto inner_col = inner_outer_pair.first;
     if (inner_col->isVirtual()) {
-      throw FailedToJoinOnVirtualColumn();
+      if (inner_outer_pairs_.size() != 1) {
+        throw HashJoinFail("Only single predicate row-id joins are currently supported.");
+      }
+      // push back an empty join column, which will prevent hash table build. the probe
+      // for a row-id hash join simply does null checks and returns the input row-id
+      auto type = inner_col->type();
+      join_columns.emplace_back(
+          JoinColumn{/*ptr=*/nullptr,
+                     /*col_chunks_buff_sz=*/0,
+                     /*num_chunks=*/1,
+                     /*num_elems=*/0,
+                     /*elem_sz=*/static_cast<size_t>(type->size())});
+      continue;
     }
     join_columns.emplace_back(fetchJoinColumn(inner_col,
                                               fragments,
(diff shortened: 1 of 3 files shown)


omniscidb/QueryEngine/JoinHashTable/PerfectJoinHashTable.cpp
```diff
@@ -971,11 +983,44 @@ llvm::Value* PerfectJoinHashTable::codegenSlot(const CompilationOptions& co,
...
+  if (val_col->isVirtual()) {
+    // short-circuit hash table load. we know the key will be the row number we want to
+    // access, so just return the key
+    CHECK(val_col_var);  // make sure this virtual column is a column
+
+    std::vector<llvm::Value*> hash_join_idx_args{
+        executor_->cgen_state_->castToTypeIn(key_lvs.front(), 64)};
+
+    const auto& query_info = getInnerQueryInfo(val_col_var).info;
+    if (query_info.getPhysicalNumTuples() == 0) {
```
ienkovich, line 997: Just curious why we can't always use `rhs_source_col_range_`. In theory, for an empty table, we should have some invalid ranges for all columns where max < min.

omniscidb/Tests/ArrowBasedExecuteTest.cpp
```diff
@@ -17849,6 +17849,58 @@ TEST_F(Select, ParseIntegerExceptions) {
...
+    c(R"(SELECT t1.x, t2.y FROM test t1 LEFT JOIN test_empty t2 ON t1.rowid = t2.rowid;)",
+      dt);
+
+    c(R"(SELECT t1.x, t2.y FROM test t1 INNER JOIN test_inner t2 ON t1.rowid = t2.rowid;)",
+      dt);
+    c(R"(SELECT t1.x, t2.y, COUNT(*) FROM test t1 INNER JOIN test_inner t2 ON t1.rowid = t2.rowid GROUP BY 1, 2 ORDER BY 1, 2;)",
+      dt);
+    c(R"(SELECT t1.x, t2.y FROM test t1 INNER JOIN test_inner t2 ON t1.rowid = t2.rowid WHERE t1.x > 7;)",
+      dt);
+    c(R"(SELECT t1.x, t2.y FROM test t1 INNER JOIN test_inner t2 ON t1.rowid = t2.rowid WHERE t1.x > 7;)",
```
ienkovich, line 17883: This query exactly matches the previous one.

ienkovich (approved):
Looks good!

omniscidb/Tests/ArrowBasedExecuteTest.cpp
alexbaden: Ah yes, this was supposed to be a filter on the inner table.

omniscidb/QueryEngine/JoinHashTable/PerfectJoinHashTable.cpp
alexbaden: Looks like it's set for `rowid` specially -- I wonder if that's worth changing? https://github.com/intel-ai/hdk/blob/main/omniscidb/QueryEngine/ExpressionRange.cpp#L547

ienkovich: Looks like a bug we might hit someday. I guess the referenced handling should go after the next if-statement, not before.

alexbaden: Hah, you're probably right. I'll try it in a separate branch.

PR #212, opened 2023-02-23
Update to gcc 12

alexbaden:


alexbaden:
This is now ready for review. 

@leshikus I made some small changes to CI to work with gcc-12. The biggest change is to using apt for cuda, instead of conda. I was also able to reduce the installed packages to just `nvcc` and its dependencies, excluding most of the cuda toolkit. That dropped install size and install time. This also allows us to use cuda 12 in CI, our preferred cuda version.

@kurapov-peter @ienkovich gcc-12 has worked fine for me in conda dev so far. Let me know if you see any issues in your envs.

ienkovich (approved):
I got the clean build locally with this branch. Thanks!

PR #213, opened 2023-02-24
ResultSet refactoring and clean-up [01/N]

ienkovich:
Remove unused QueryMemoryDescriptor::allow_multifrag_.

I'm starting my work on `ResultSet` refactoring. This is preparation work to make `ResultSet` and related structures independent of the `QueryEngine` library. The ultimate goal is to extract structures describing memory layout and transform `ResultSet` into a class that holds data and provides access to it (and leaving the rest like the reduction in QueryEngine). Basically, that would be kind of an analog of `arrow::Table` but covering all our possible output formats. All related code would go into the separate library which later will be used to create `ResultSetStorage` to unify access to data for the QueryEngine and make `ResultSet` a first-class citizen, enabling queries over ResultSets (would be very useful for Modin and Python API in general where we have to convert and re-import query result to continue processing).

The first part is a trivial clean-up of an unused field I noticed reading through ResultSet-related structures. I'll try to make multiple small PRs in this series rather that a single mega PR.

### DIFF
--- a/omniscidb/QueryEngine/Descriptors/QueryMemoryDescriptor.cpp
+++ b/omniscidb/QueryEngine/Descriptors/QueryMemoryDescriptor.cpp
@@ -426,7 +426,6 @@ QueryMemoryDescriptor::QueryMemoryDescriptor(
     const bool must_use_baseline_sort,
     const bool use_streaming_top_n)
     : executor_(executor)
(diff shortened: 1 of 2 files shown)


alexbaden (approved):
Weird that this isn't used, but likely fine.

PR #216, opened 2023-02-24
ResultSet refactoring and clean-up [04/N]

ienkovich:
`ExecutorDeviceType` is a very common enum actually, so move to the Shared lib.

How about renaming it to simply `DeviceType`?


kurapov-peter (approved):
I don't mind renaming

PR #220, opened 2023-02-28
ResultSet refactoring and clean-up [07/N]

ienkovich:
Get `BufferProvider` from `DataMgr` in `Executor` and `ResultSet`.

This is just to avoid unnecessary fields and args and make it more explicit that `BufferProvider` is always provided by `DataMgr`.


kurapov-peter (approved):
LGTM

PR #224, opened 2023-03-01
Remove Docs CI and add pytest

### LINKED ISSUE #223: CI doesn't report broken PyHDK tests
About a month ago `python/tests/test_pyhdk_bindings.py::TestArrowStorage::test_import_arrow_table` test was broken and CI didn't report it.

alexbaden:
docs is no longer used and was causing failing tests on main. 

I added a new pytest only run which does not include modin. The modin CI script still needs some work, since it isn't running tests properly, but I wanted to have CI that wouldn't have any possibility of using the pyhdk version from modin. 

Closes #223


alexbaden:
Let's just merge it so we have pytest coverage. The other tests do not build hdk from root. Modin test is different. We can improve both after we have the coverage running consistently again.

ienkovich (approved):
Since there are no build artifacts to reuse right now, let's use this solution and unify later.

PR #227, opened 2023-03-01
ResultSet refactoring and clean-up [08/N]

ienkovich:
Remove Executor usages from QueryMemoryDescriptor.

This is achieved by two changes:
1. Use `DataMgr` instead of `Executor`
2. Move a big part of the `QueryMemoryDescriptor` code to `MemoryLayoutBuilder` and `QueryExecutionContext`. The goal is to make `QueryMemoryDescriptor` just a descriptor that shouldn't know about entities using and creating it.


kurapov-peter (approved):
LGTM

PR #229, opened 2023-03-01
Add a workflow for Intel GPU enabling tests

kurapov-peter:
Using all the changes and the latest driver version helped. This now passes in CI. Hooray.


leshikus (approved):
Each RUN command creates a separate file system layer which is independently cached; this may be difficult to debug

I suggest merging multiple RUN commands into one; otherwise looks good

PR #231, opened 2023-03-02
ResultSet refactoring and clean-up [10/N]

ienkovich:
`ResultSetStorage` class has a lot of methods related to reduction. All implementations are actually in `ResultSetReduction.cpp`. To make `ResultSetStorage` class simpler and be unaware of `Executor`, `ReductionCode`, etc. I moved this code into a new `ResultSetReduction` class. Later `ResultSetStorage` will become a part of `ResultSet` library and reduction related code will be kept in `QueryEngine`.


kurapov-peter (approved):
LGTM!

PR #233, opened 2023-03-02
ResultSet refactoring and clean-up [12/N]

ienkovich:
`ResultSetStorage` uses decoders from the runtime which makes it dependent on the `QueryEngine`. I added new decoders to be used in statically compiled code only to avoid this dependency.

Also, moved empty key declarations to Shared lib.


kurapov-peter (approved):
Do you want to merge this or move to a separate library firts?

PR #238, opened 2023-03-03
Fixed __restrict__ definition on windows

gshimansky:
Use cross-platform define for __restrict__ keyword


gshimansky:
PR closed and #257 opened instead

PR #239, opened 2023-03-03
Added dependency on Logger and OSDependent libraries

gshimansky:
Added explicit dependencies from Logger and OSDependent libraries. They are required on windows because unresolved symbols are not allowed.

### DIFF
--- a/omniscidb/OSDependent/Windows/CMakeLists.txt
+++ b/omniscidb/OSDependent/Windows/CMakeLists.txt
@@ -1,1 +1,4 @@
 add_library(OSDependent ${OSDEPENDENT_SOURCE_FILES})
+set_target_properties(OSDependent PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/..)
+
+target_link_libraries(OSDependent Logger)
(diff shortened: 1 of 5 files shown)


omniscidb/OSDependent/Windows/CMakeLists.txt
```diff
@@ -1 +1,4 @@
 add_library(OSDependent ${OSDEPENDENT_SOURCE_FILES})
+set_target_properties(OSDependent PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/..)
+
+target_link_libraries(OSDependent Logger)
```
kurapov-peter, line 4: IMO, Logger should be a private dependency here as well as in the device managers. What made you decide otherwise?

gshimansky: I haven't made any decisions and simply copied these lines from `Unix/CMakeLists.txt`.

kurapov-peter: I guess it doesn't hurt too much, we'll deal with it separately.

PR #240, opened 2023-03-03
Fixed #warning preprocessor directive

gshimansky:
Added condition to use of `#warning` preprocessor directive

### DIFF
--- a/omniscidb/QueryEngine/TargetExprBuilder.cpp
+++ b/omniscidb/QueryEngine/TargetExprBuilder.cpp
@@ -28,6 +28,8 @@
 #include "MaxwellCodegenPatch.h"
 #include "OutputBufferInitialization.h"
 
+#include <boost/config/pragma_message.hpp>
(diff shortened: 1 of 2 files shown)


alexbaden (approved):
Looks fine for now. Could we make a macro for generating compiler warnings for future use, so we don't have to #ifdef?

gshimansky:
@alexbaden Is there a header file in HDK that is included in all other places in the code where we could define such macro?

alexbaden:
No, usually we would create a new header file in `Shared` and define the macro there. It can then be included where needed. See, for example, `Shared/checked_alloc.h` (not a macro but a free function that is used in a lot of different places).

gshimansky:
@alexbaden There is [BOOST_PRAGMA_MESSAG](https://www.boost.org/doc/libs/1_81_0/boost/config/pragma_message.hpp). What do you think about using it instead?

PR #243, opened 2023-03-03
Updated windows conda environment

gshimansky:
Match versions in linux environment


gshimansky:
PR closed and #256 opened instead

PR #245, opened 2023-03-03
Resolve cicular dependency from Analyzer to QueryEngine

gshimansky:
Use an abstract class with virtual function to resolve linking circular dependency between Analyzer and QueryEngine


ienkovich (changes requested):
Since we are trying to break the dependency on `QueryEngine`, could you please replace all usages of `Executor` in this file, not just a single occurrence, and also remove `QueryEngine/Execute.h` include?

gshimansky:
An extra push had to be done to satisfy style checker. Please merge now.

PR #248, opened 2023-03-03
Enabled table functions generation scripts on windows

gshimansky:
Remove conditions that prevent running table generation scripts on windows


alexbaden (approved):
I am surprised this built. Were changes required elsewhere to make these work?

gshimansky:
The only necessary change was to remove all of `if(NOT MSVC)` conditions. Initially I left some of them in place and build failed. Possibly it doesn't build on windows without some of other patches that were done before it, but it shouldn't affect Linux anyway.

PR #250, opened 2023-03-03
Added dllimport/dllexport where necessary

gshimansky:
Added correct dllimport and dllexport attributes necessary to link libraries and executables on windows

### DIFF
--- a/omniscidb/QueryEngine/Execute.cpp
+++ b/omniscidb/QueryEngine/Execute.cpp
@@ -14,8 +14,11 @@
  * limitations under the License.
  */
 
-#include "QueryEngine/Execute.h"
+#include "Shared/funcannotations.h"
+#define QUERYENGINE_EXPORT RUNTIME_EXPORT
+
 #include "CardinalityEstimator.h"
+#include "QueryEngine/Execute.h"
 
 #include <llvm/Transforms/Utils/BasicBlockUtils.h>
 #include <boost/filesystem/operations.hpp>
--- a/omniscidb/Tests/ResultSetBaselineRadixSortTest.cpp
+++ b/omniscidb/Tests/ResultSetBaselineRadixSortTest.cpp
@@ -40,7 +40,7 @@ extern std::unique_ptr<CudaMgr_Namespace::CudaMgr> g_cuda_mgr;
 #include <numeric>
 #include <random>
 
-extern bool g_is_test_env;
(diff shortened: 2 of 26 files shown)


kurapov-peter (commented):
Can we have a single header for these definitions? In Shared for instance.

gshimansky:
> Can we have a single header for these definitions? In Shared for instance.

Windows specific is that you need to declare the same function/variable differently depending on whether you compile a DLL that contains this function/variable or you compile a code that links with this DLL.
https://wiki.tcl-lang.org/page/DLLEXPORT+and+DLLIMPORT

- When you compile DLL itself, function/variable has to be declared as `__declspec(dllexport)`. This is optional because we use cmake mode `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` that makes all DLL symbols as dllexport. But we cannot mark them as `dllimport` in the header because this leads to a linker messages about inconsistency.
- When compiling code external to the DLL, all symbols in that DLLs API headers should be marked as `dllimport`. This is mandatory and linker doesn't find those symbols if they are not explicitly marked.

This is the reason for conditional compilation in DLL headers and also different constants for each DLL.

omniscidb/Tests/ResultSetBaselineRadixSortTest.cpp
```diff
@@ -39,7 +39,10 @@ extern std::unique_ptr<CudaMgr_Namespace::CudaMgr> g_cuda_mgr;
 #include <numeric>
 #include <random>
 
-extern bool g_is_test_env;
+#ifdef _WIN32
```
ienkovich, line 42: Do we have to add this to all extern declarations? Maybe make EXTERN macro then which would work for both Linux and Windows?

alexbaden:
@gshimansky what is the status of this PR?

omniscidb/QueryEngine/Execute.cpp
```diff
@@ -14,8 +14,10 @@
  * limitations under the License.
  */
 
-#include "QueryEngine/Execute.h"
+#include "Shared/funcannotations.h"
+#define QUERYENGINE_EXPORT RUNTIME_EXPORT
```
ienkovich, line 18: I'd add an empty line after such defines to not hide them within a bunch of includes. Otherwise looks good.

ienkovich (approved):

PR #251, opened 2023-03-03
Removed field name from initializers because MSVC thinks it is C++20

gshimansky:
MSVC on windows doesn't allow specifying field names like this. It complains that this is a part of C++20 but enabling C++20 breaks a lot of code.


alexbaden (approved):
Sad that we can't still use this. I wonder if there is `clang-tidy` or `clang-format` rule to disallow use in CI (until we get Windows CI going).

alexbaden:
Approved too soon -- @gshimansky looks like it needs a `clang-format` run. If you want I can run it and push an update to the branch.

PR #258, opened 2023-03-04
ResultSet refactoring and clean-up [13/N]

ienkovich:
Move the cardinality expression to IR. Also, move related global flags to `Config`. Also, a single estimator implementation is enough, for now, the large version adds nothing new.

It reveals some implementation details to IR level, but otherwise, we cannot interpret `ResultSet` with estimator results without a dependency on `QueryEngine`. I think having estimator expressions in IR is better.

### DIFF
--- a/omniscidb/ConfigBuilder/ConfigBuilder.cpp
+++ b/omniscidb/ConfigBuilder/ConfigBuilder.cpp
@@ -166,6 +166,15 @@ bool ConfigBuilder::parseCommandLineArgs(int argc,
       po::value<size_t>(&config_->exec.group_by.baseline_threshold)
           ->default_value(config_->exec.group_by.baseline_threshold),
       "Prefer baseline hash if number of entries exceeds this threshold.");
+  opt_desc.add_options()("large-ndv-threshold",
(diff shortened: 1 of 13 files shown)


kurapov-peter:
Is this a temporary solution? The cardinality estimation should probably live somewhere in the `QueryEngine`.

ienkovich:
No, this is not a temporary solution. The estimator is an IR expression, and I move this expression to the IR library, I don't move related runtime. Estimator expression is required to interpret ResultSet data, so code is moved to break ResultSet dependency on QueryEngine.

This probably can be reorganized in some other way. E. g. move all required info to QueryMemoryDescriptor and drop the estimator expression from ResultSet. But still, I don't see why we shouldn't keep expression declarations in IR library.

kurapov-peter:
Why is the estimator an IR expression in the first place? I.e. why would a use want to create an estimator expression?

ienkovich:
Estimator is an expression because we mostly use IR expressions to build execution units. It doesn't have to be an expression actually, but it still should be some structure available outside of the QueryEngine to interpret execution results.

ienkovich:
@kurapov-peter Any thoughts on this one?

PR #259, opened 2023-03-05
Add a release artifact

leshikus:
Subj


pbchekin (approved):
LGTM.
The only note that every artifact then dynamically archived to zip (when downloading? see https://github.com/actions/upload-artifact#zipped-artifact-downloads), so there is double archiving. The question is does it make sense to upload .tar only without gzipping it because later it will be zipped by GitHub. For a sake of simplicity I am proposing to keep it as is: uploading .tag.gz).

leshikus:
@pbchekin that's correct; though there is a problem that this zip archiving implementation was very slow for directories; for a small number of files it works

PR #266, opened 2023-03-08
Fix python install

leshikus:
Subj


omniscidb/scripts/conda/build.sh
```diff
@@ -5,5 +5,5 @@ cd $(dirname "$0")/../../..
 
 test -z "$CC" || export CUDAFLAGS="--compiler-bindir $CC"
 cmake -B build -S . $@
-cmake --build build --parallel 2
+cmake --build build --parallel `nproc`
```
kurapov-peter, line 8: Didn't we set it to 2 to satisfy default github runners requirements?

leshikus: 1) yes, we did,
2) for CI this should work the same way; if someone wants to use this script locally, he may want to have more processors involved
3) a year ago it was not possible to use too many processors, there were some errors related to job parallelization

ienkovich (approved):
LGTM

PR #267, opened 2023-03-08
ResultSet refactoring and clean-up [15/N]

ienkovich:
Move `ExtractFromTime` runtime to `Utils` to break `ResultSet` dependency on `QueryEngine` runtime. This is a temporary hack, the proper solution will be to move runtime into a separate library. But this seems to be a non-trivial task, so we will do it separately.


aregm (approved):
As Alex is vacationing and we all envy him very much, I approve of this mere renaming.

PR #278, opened 2023-03-10
Enable sanitizers in root CMakeLists.txt

kurapov-peter:
If built from root, our libraries don't link with sanitizers now. Here's a fix.


leshikus (approved):
LGTM

PR #281, opened 2023-03-13
Speed up pytest github actions job

alexbaden:
Tentative fix for the `pytest` failures.


leshikus (commented):
I would suggest adding the pytest.yml as a subtask to test.yml (after the problem with intermittent failures is fixed), this would also remove extra rebuild

alexbaden:
Job was fixed here: https://github.com/intel-ai/hdk/pull/273
I dropped that fix from this branch, but I think the changes to use the conda cache are worth keeping. It improves build time from ~45 min to ~25 min. 
Making it a subtask is a good idea, but I think we should handle that separately.

ienkovich (approved):

PR #284, opened 2023-03-14
Remove unused features and unused code

alexbaden:
See commit names for an exhaustive list. Given each unit test is 400MB, the removal of three unit tests alone saves over 1GB of build space. The other removals were targeted to improve layering (query hints ends up scattering configs across a bunch of function signatures), reduce signature size, and/or reduce branch complexity, particularly for features that were always on (for several years).


ienkovich (approved):
I was thinking about keeping table functions as the only way to process a query colwise. But you and Petr really want them to go :) Well, I guess it will be better to not have it on IR level anyway.

Nice clean-up! Thanks!

P.S. Do we still need `hdk::ir::Type::kColumn` and `hdk::ir::Type::kColumnList`?

alexbaden:
We can probably drop `kColumn` and `kColumnList`. I initially went farther in dropping table functions but that ended up breaking varlen extension functions (like `array_append`) so I had to backtrack.

PR #285, opened 2023-03-15
Use ResultSet print API in StandaloneQueryRunner

alexbaden:


omniscidb/Tests/StandaloneQueryRunner.cpp
```diff
@@ -26,7 +26,11 @@ using namespace TestHelpers;
 namespace {
 
 const size_t g_num_rows{10};
+#ifdef HAVE_CUDA
```
kurapov-peter, line 29: This should check both L0 and CUDA.

kurapov-peter (changes requested):

kurapov-peter (approved):

PR #287, opened 2023-03-15
Purge gcc

leshikus:
Subj

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Free space
+        run: |
+          sudo rm -rf /usr/local/lib/android/sdk /opt/hostedtoolcache /usr/share/swift /opt/az
+          sudo apt-get purge -y gcc llvm-12 llvm-13 llvm-14
+
       - name: Set env context
         run: |
           echo RUN_STAMP=${{ runner.os }}-${{ inputs.name }} >>$GITHUB_ENV


alexbaden (changes requested):
The fix is fine in general, but I am wondering why this is happening now. Presumably we didn't have a problem only reading from conda paths before. What changed?

My concern is that we could be masking a problem that will appear in certain conda dev environments, and we may end up using the paths that are outside the conda env when we think we are in fact using everything from within the conda env.

leshikus:
Maybe this was caused by github image update which added some incompatibility between gcc-11 and gcc-12 headers

alexbaden:
My latest push to main didn't fail: https://github.com/intel-ai/hdk/actions/runs/4432705489/jobs/7777714872

I noticed the conda cache key is the same, except for the date:
```
job that failed, plus two retry attempts (also failed):
Cache restored from key: Linux-cpu-conda-3554347203db3bd910f7b11f421ff828e904d2a0e99b068216c374c615068a8f-20230315
Cache restored from key: Linux-cpu-conda-3554347203db3bd910f7b11f421ff828e904d2a0e99b068216c374c615068a8f-20230315
Cache restored from key: Linux-cpu-conda-3554347203db3bd910f7b11f421ff828e904d2a0e99b068216c374c615068a8f-20230315

main job that passed:
Cache restored from key: Linux-cpu-conda-3554347203db3bd910f7b11f421ff828e904d2a0e99b068216c374c615068a8f-20230316
```
Maybe a bad conda cache file?

leshikus:
I believe you are correct; I still find this change - making a test container look more like a build container, - useful

leshikus:
I've checked the last passed build, it used Cache restored from key: Linux-cpu-conda-3554347203db3bd910f7b11f421ff828e904d2a0e99b068216c374c615068a8f-20230314
https://github.com/intel-ai/hdk/actions/runs/4422314475/jobs/7753992525

this also supports your version

PR #289, opened 2023-03-16
Fix misc compiler warnings 

alexbaden:


omniscidb/QueryEngine/EquiJoinCondition.h
```diff
@@ -19,6 +19,8 @@
 #include <list>
 #include <memory>
 
+#include "IR/Expr.h"
+
 namespace hdk::ir {
```
ienkovich, line 24: No need for these declarations then.

ienkovich (approved):

PR #226, opened 2023-03-01
Folly enabled by default

kurapov-peter:
The failures we saw were cause by cmake picking up dependencies via the 'module path'. Also, Folly searches for its dependencies and overwrites `Boost_LIBRARIES` which caused linking issues in the logger and hence almost everywhere. I removed the find-modules altogether since they are outdated anyway.

The version of Folly is the same as we have in conda.

PR #293, opened 2023-03-18
Enable folly 

alexbaden:
This PR is almost identical to #226, but includes one additional commit which propagates the folly compiler definition to the cython code. The previous segfault occurred because cython code was using a `std::shared_timed_mutex`, whereas the compiled libraries were using `folly::mutex`. 

This does bring up an interesting issue which is the need to propagate compiler definitions to cython more generally. I just did a one-off to get folly working, but I am interested in ways to propagate the set definitions using cmake variables (as a followup). 

Also, I found a cython flag to enable gdb. On ubuntu using the conda environment, this gave me enough debug symbols (with regular gdb) to get a stacktrace. I turned the flag off but left it in `setup.py.in` for future use. 

Supersedes #226


ienkovich:
> The previous segfault occurred because cython code was using a `std::shared_timed_mutex`, whereas the compiled libraries were using `folly::mutex`.

This is a good catch. We also have other similar conditional declarations, e.g. for threading. We should get all definitions from `COMPILE_DEFINITIONS` property to be more safe here. I created #297 for that.

ienkovich (approved):

CMakeLists.txt
```diff
@@ -49,6 +49,10 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
             "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
 endif ()
 
+# External Dependencies
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake_modules")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/omniscidb/cmake/Modules")
```
kurapov-peter, line 54: I wish we could avoid using any legacy cmake stuff

kurapov-peter (approved):
Thanks! There're some minor cmake comments

PR #298, opened 2023-03-21
Enable pull request check

leshikus:
Subj


leshikus:
There was a request from Petr to enable pull_request checks, I believe this is a correct way to go

kurapov-peter (approved):
Thanks!!

PR #312, opened 2023-03-23
Remove freeze instruction in Backend.cpp

lmontigny:


omniscidb/QueryEngine/Compiler/Backend.cpp
```diff
@@ -981,6 +981,23 @@ std::shared_ptr<L0CompilationContext> L0Backend::generateNativeGPUCode(
   pass_manager_builder.populateModulePassManager(PM);
   compiler::optimize_ir(func, module, PM, live_funcs, false /*smem_used*/, co);
 
+  // Remove the remaining freeze instruction after the optimization
+  std::vector<llvm::Instruction*> ToErase;
```
alexbaden, line 985: Can we change the variable case here to match existing style? https://google.github.io/styleguide/cppguide.html

alexbaden (changes requested):

omniscidb/QueryEngine/Compiler/Backend.cpp
lmontigny: Sure, I took it directly from the SPRIV LLVM Translator

alexbaden (approved):

PR #326, opened 2023-03-27
Linear regression extrapolation and DwarfBench API fixes

bagrorg:
This pull request introduces a `Linear regression model` to extrapolate data for `Cost model`.

It also brings some fixes related to `DwarBench` API changes.


omniscidb/Tests/CostModel/CostModelTest.cpp
```diff
@@ -32,6 +34,25 @@ class DataSourceTest : public DataSource {
   }
 };
 
+bool doubleEquals(double x, double y) {
```
kurapov-peter, line 37: Please use gtest's `ASSERT_NEAR`. http://google.github.io/googletest/reference/assertions.html#EXPECT_NEAR

omniscidb/Tests/CostModel/CostModelTest.cpp
```diff
@@ -32,6 +34,25 @@ class DataSourceTest : public DataSource {
...
+
+class LinearRegressionTest : public LinearRegression {
+ public:
+  LinearRegressionTest(std::vector<Detail::Measurement>& measurement)
+      : LinearRegression(measurement) {}
+
+  void equalsWs(arma::vec wExpected) {
+    ASSERT_EQ(w.size(), wExpected.size());
+
+    for (size_t i = 0; i < w.size(); i++) {
```
kurapov-peter, line 50: For vector comparisons and such, please consult with `TestHelpers.h` and `ArrowTestHelpers.h` first.

omniscidb/QueryEngine/CostModel/ExtrapolationModels/LinearRegression.h
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  size_t getExtrapolatedData(size_t bytes) override;
+
+ protected:
+  void buildRegressionCoefficients();
+  arma::mat buildFeaturesMatrix();
+  arma::vec buildTargets();
+
+  arma::vec w;
```
kurapov-peter, line 39: We usually use underscores for private or protected members. When in doubt, consult with https://google.github.io/styleguide/cppguide.html. This applies to the whole PR.

omniscidb/QueryEngine/CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ list(APPEND query_engine_source_files
         CostModel/CostModel.cpp
         CostModel/DataSources/EmptyDataSource.cpp
         CostModel/ExtrapolationModels/LinearExtrapolation.cpp
+        CostModel/ExtrapolationModels/LinearRegression.cpp
```
kurapov-peter, line 136: I think we should make the cost model live in a separate library. Here you're unconditionally adding the sources to the QueryEngine library, thus leaking the dependency.

kurapov-peter (changes requested):

omniscidb/Tests/CostModel/CostModelTest.cpp
bagrorg: Should i add comparison function for armadillo vectors?

omniscidb/QueryEngine/CMakeLists.txt
bagrorg: Then we also want to hide `armadillo` and `DwarfBench` dependencies from headers, am I right?

omniscidb/Tests/CostModel/CostModelTest.cpp
kurapov-peter: This is FYI, since you have a single smoke test there's no need for it to be in the header.

omniscidb/QueryEngine/CostModel/DataSources/DataSource.h
```diff
@@ -15,6 +15,7 @@
 
 #include "QueryEngine/CostModel/Measurements.h"
 
+#include <set>
```
kurapov-peter, line 18: Looks unnecessary

omniscidb/QueryEngine/CostModel/DataSources/DwarfBench.h
```diff
@@ -21,34 +21,31 @@
...
 
-  DwarfBench::Dwarf convertToDwarf(AnalyticalTemplate templ);
-  DwarfBench::DeviceType convertDeviceType(ExecutorDeviceType device);
-
-  std::vector<Detail::Measurement> convertMeasurement(
-      const std::vector<DwarfBench::Measurement> measurements);
-
-  DwarfBench::DwarfBench db;
+  struct PrivateImpl;
+  PrivateImpl* pimpl_;
```
kurapov-peter, line 48: Any reason not to use a `unique_ptr`?

omniscidb/QueryEngine/CostModel/ExtrapolationModels/LinearRegression.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "LinearRegression.h"
+#include <armadillo>
+
+namespace costmodel {
+
+struct LinearRegression::PrivateImpl {
+  arma::mat buildFeaturesMatrix(const std::vector<Detail::Measurement>& measurement);
+  arma::vec buildTargets(const std::vector<Detail::Measurement>& measurement);
+
+  arma::vec w_;
```
kurapov-peter, line 23: The name implies the member is private which is a bit confusing. I'd change it to something like `weights`.

kurapov-peter (commented):
Overall looks good, a couple more comments.

omniscidb/QueryEngine/CostModel/ExtrapolationModels/LinearRegression.cpp
```diff
@@ -0,0 +1,76 @@
...
+  buildRegressionCoefficients();
+}
+
+LinearRegression::~LinearRegression() {
+  delete pimpl_;
+}
+
+size_t LinearRegression::getExtrapolatedData(size_t bytes) {
+  arma::vec x = {1.0, (double)bytes};
+  return (size_t)arma::dot(x, pimpl_->w_);
```
akroviakov, line 42: Would be great to avoid such C-style casts.

omniscidb/QueryEngine/CostModel/DataSources/DwarfBench.h
bagrorg: I thought that because of `static_assert` in `unique_ptr` this is impossible, it turned out that it was necessary to explicitly prescribe a destructor

kurapov-peter (changes requested):
The git history is messed up, could you please clean it up @bagrorg?

PR #328, opened 2023-03-28
Fix python version to 3.9

leshikus:
This is a workaround for https://github.com/intel-ai/hdk/issues/332 (it fixes development env; runtime env is still needed)

### DIFF
--- a/.github/workflows/modin.yml
+++ b/.github/workflows/modin.yml
@@ -29,7 +29,6 @@ jobs:
         run: |
           conda update conda
           conda env update -f omniscidb/scripts/mapd-deps-conda-dev-env.yml
-          conda install -n ${{ env.CONDA_ENV }} -c conda-forge python=3.8
 
       - name: Restore Maven cache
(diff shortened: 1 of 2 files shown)


Garra1980:
try 
- python 3.9.16
in yml file instead just
- python 3.9

alexbaden:
Shouldn't we use python `3.9.*` or equivalent so we will get security updates, etc without having to manually bump the minor version each time?

Garra1980:
> Shouldn't we use python `3.9.*` or equivalent so we will get security updates, etc without having to manually bump the minor version each time?

yes, I think we should try this

kurapov-peter:
What's the point of setting the version for our dev envs anyway?

leshikus:
fixed to 3.9.*

Garra1980:
> What's the point of setting the version for our dev envs anyway?

to sync with our benchmarking envs at least

alexbaden (approved):

PR #336, opened 2023-03-29
Fix python version to 3.9

leshikus:
Subj with 3.9.*


alexbaden (approved):

alexbaden (changes requested):
Looks like github actions is down... let's wait until it is back up so we can re-run CI with this change.

PR #338, opened 2023-03-30
Use __atomic_compare_exchange_n for CPU CAS atomics where intrinsics …

alexbaden:
…are used.

It turns out that the x86 assembly for the two intrinsics is the same:

```
int mapd_cas_test(int64_t* matching_group) {
  int64_t invalid_slot_val = std::numeric_limits<int64_t>::max();
int64_t write_pending = invalid_slot_val - 1;
 auto res = mapd_cas(matching_group, invalid_slot_val, write_pending);
 if (res != write_pending) {
    return -1;
  }
  return 0;
}
int64_t* cas_cst_test(int64_t* row_ptr, int64_t* key ) {
    int64_t empty_key = std::numeric_limits<int64_t>::max();
    int64_t write_pending = empty_key - 1;
  const bool success = cas_cst(row_ptr, &empty_key, write_pending);
  if (success) {
    store_cst(row_ptr, *key);
    return reinterpret_cast<int64_t*>(row_ptr + 1);
  }
}
```

(where `mapd_cas` calls `__sync_val` and `cas_cst` calls `__atomic`)

``` 
mapd_cas_test(long*):                    # @mapd_cas_test(long*)
        movabs  rax, 9223372036854775807
        movabs  rdx, 9223372036854775806
        lock            cmpxchg qword ptr [rdi], rdx
        xor     ecx, ecx
        cmp     rax, rdx
        setne   cl
        neg     ecx
        mov     eax, ecx
        ret
cas_cst_test(long*, long*):                   # @cas_cst_test(long*, long*)
        movabs  rax, 9223372036854775807
        movabs  rcx, 9223372036854775806
        lock            cmpxchg qword ptr [rdi], rcx
        jne     .LBB1_2
        mov     rax, qword ptr [rsi]
        xchg    qword ptr [rdi], rax
        add     rdi, 8
        mov     rax, rdi
        ret
```

The `cmpxchg` instructions sets the appropriate flags depending on what the compare result was, but in the `__sync_val` case those flags are never read. Uniformly using `__atomic` results in fewer instructions, and `__sync_val` is deprecated anyway.

@gshimansky I left all atomic names the same to reduce conflicts w/ Windows. We can make Windows work, then do the renaming, or do the renaming as part of the Windows work, as we no longer need `mapd_cas` and `cas_crt`.

### DIFF
--- a/omniscidb/QueryEngine/JoinHashTable/Runtime/HashJoinRuntime.cpp
+++ b/omniscidb/QueryEngine/JoinHashTable/Runtime/HashJoinRuntime.cpp
@@ -215,16 +215,50 @@ void SUFFIX(init_hash_join_buff_tbb)(int32_t* groups_buffer,
 #endif  // #ifndef __CUDACC__
 
 #ifdef __CUDACC__
-#define mapd_cas(address, compare, val) atomicCAS(address, compare, val)
+#define hdk_cas(address, compare, val) atomicCAS(address, compare, val)
 #elif defined(_MSC_VER)
-#define mapd_cas(address, compare, val)                                 \
-  InterlockedCompareExchange(reinterpret_cast<volatile long*>(address), \
-                             static_cast<long>(val),                    \
-                             static_cast<long>(compare))
+#define hdk_cas(ptr, expected, desired) template <typename T>
+template <typename T>
+bool hdk_cas(T* ptr, T* expected, T desired) {
+  if constexpr (sizeof(T) == 4) {
+    return InterlockedCompareExchange(reinterpret_cast<volatile long*>(ptr),
+                                      static_cast<long>(desired),
+                                      static_cast<long>(*expected)) ==
+           static_cast<long>(*expected);
+  } else if constexpr (sizeof(T) == 8) {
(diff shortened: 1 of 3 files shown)


ienkovich (approved):
Looks good.

When we have a working Windows version, I'd also propose moving these definitions into a separate header in the `Shared` lib. Also, make them template functions. This would provide better type checks (because we use `interpret_cast` on Windows) and allow immediate values for the second operand in `mapd_cas`.

gshimansky:
I wonder what is the difference between `mapd_cas` and `cas_cst` after the change? Can we use just one CAS operation in `HashJoinRuntime.cpp`?

alexbaden:
I can unify the two implementations, that is certainly cleaner. I didn't want to do it if it would conflict with your patch, but let me go ahead and do that (and fix the issue here, I think I just got one CAS backwards).

alexbaden:
I think the stack smashing test failures were the result of type mismatches in `__atomic_compare_exchange_n` -- because we're now comparing two pointers, I assume the types must match. I could not reproduce locally so I am going to re-run a few CI runs here to make sure there is not some underlying transient issue.

PR #341, opened 2023-03-31
Revert pr266

leshikus:
Subj


Devjiu:
It's better to add a link to the original pr in such cases. like this: #266

Devjiu (approved):
I agree with this revert if the tests are passed.

PR #345, opened 2023-04-03
[CI] Run style check in parallel

Devjiu:
This PR allow to run formatting in parallel with build and all other jobs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -87,3 +87,42 @@ jobs:
 
   pytest:
     uses: ./.github/workflows/pytest.yml
+
+  style-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set env context
+        run: |
+          echo RUN_STAMP=${{ runner.os }}-${{ inputs.name }} >>$GITHUB_ENV
+          echo CONDA_ENV_PATH=$CONDA/envs/omnisci-dev >>$GITHUB_ENV
+          echo "DATE=$(date +'%Y%m%d')" >> $GITHUB_ENV
+
+      - name: Restore Conda env cache
+        id: conda-cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ${{ env.CONDA_ENV_PATH }}
+          key: ${{ env.RUN_STAMP }}-conda-${{ hashFiles('omniscidb/scripts/mapd-deps-conda-dev-env.yml') }}-${{ env.DATE }}
+          restore-keys: |
+            ${{ env.RUN_STAMP }}-conda-${{ hashFiles('omniscidb/scripts/mapd-deps-conda-dev-env.yml') }}-
+
+      - name: Update Conda env
+        if: steps.conda-cache.cache-hit != 'true'
+        run: |
+          conda update conda
+          conda env update -f omniscidb/scripts/mapd-deps-conda-dev-env.yml
+
+      - name: C++ Style check
+        run: |
+          source /usr/share/miniconda/etc/profile.d/conda.sh
+          conda activate omnisci-dev
+          find . -name build -prune -or -name ThirdParty -prune -or -name third_party -prune -or \( -name '*.cpp' -print -or -name '*.h' -print \) | xargs clang-format --dry-run --Werror -style=file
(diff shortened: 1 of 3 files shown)


Devjiu:
https://github.com/intel-ai/hdk/actions/runs/4598011653 - example of cancel via fail of style check.

Devjiu:
> What happens if we update to LLVM 16?

In both cases, we will have to update the workflow. If the clang format changes the formatting rules and you want to have the cutting-edge clang format in both cases (currently used and suggested in this pr), we need a pr with reformatting all files.

leshikus (commented):
The point of placing style check behind was we need conda env first, so the check will use clang tools from the conda env

this one uses system tools, they may be of a different version or just absent

Devjiu:
> The point of placing style check behind was we need conda env first, so the check will use cland tools from conda env
> 
> this one uses system tools, they may be of a different version or just absent

Ok, there are 2 approaches to avoid missing clang-format in system. 
1. restore conda cache and use it from cache. 
2. add apt install before.

We already have several parameters, that are global for whole project (e.g. llvm version, cuda version) Currently it's hardcoded in several places. (conda env yaml and Dockerfile) 

From my point of view clang-format version also is global parameter. Changes in this parameters requires reformatting of whole project. (clang-format different versions detects different warns/errors)

Devjiu:
Removed workflow cancellation requested by @ienkoovich. Because the repository from the project's global dependency versions did not reach consensus, I used the currently cached conda env. The llvm/cuda dependency is currently listed in the Dockerfile and in the conda env. It was decided that the problem is not critical, so we'll just leave it as it is until we face it face to face.

Devjiu:
As it connected with llvm update process, @alexbaden PTAL and if you have any comments or issues with update process please share your opinion.

ienkovich:
> From my point of view clang-format version also is global parameter. Changes in this parameters requires reformatting of whole project. (clang-format different versions detects different warns/errors)

In our development, we use clang-format from conda env. I don't see any need for a new parameter, we had no troubles with formatting previously when moved to the newer LLVM version.

It's probably nice to have style check in parallel with the build but looks like PR introduces some code duplications. We definitely don't need two style checks. I'm not sure if conda env restore code can be re-used instead of copied.

Devjiu:
> > From my point of view clang-format version also is global parameter. Changes in this parameters requires reformatting of whole project. (clang-format different versions detects different warns/errors)
> 
> In our development, we use clang-format from conda env. I don't see any need for a new parameter, we had no troubles with formatting previously when moved to the newer LLVM version.
> 
> It's probably nice to have style check in parallel with the build but looks like PR introduces some code duplications. We definitely don't need two style checks. I'm not sure if conda env restore code can be re-used instead of copied.

In further updates for cache maintenance will be suggested  to use custom gh action. We currently have code duplication in current approach. (conda-cache - build.yml and pytest.yml , m2 cache - build.yml, modin.yml, pytest.yml)

ienkovich:
You still have two separate style check actions now.

IMO it would be more useful to add style check cmd to a pre-commit hook and leave CI style check as is. Adding python formatting would be nice though.

Devjiu:
> You still have two separate style check actions now.
> 
> IMO it would be more useful to add style check cmd to a pre-commit hook and leave CI style check as is. Adding python formatting would be nice though.

Action in terms of github is something, that located in action.yml and is a simple bunch of consecutive steps. It's similar to function to isolate common algorithm for different input parameters, but not so big as workflow, that can have several parallel jobs inside. (example of cache action from my previous approach - https://github.com/intel-ai/hdk/blob/3b921ab874f719ae7af19f1c556e58a621f20edb/actions/cache/action.yml )

I removed style section from test.yml. Style test don't have any logical dependency on build, so I am removing it. If we don't care about resources - second time cache check shouldn't be an issue. Similar code duplication already happened in several jobs. Current change only removes false dependency.

Devjiu:
Currently `conda update` is used to save env and validate it on build step. After success of build step it's updated. Initial idea was to make update only one time. This pr violates this logic, so at first `conda update` and env caching should be done during nightly. Further jobs (pr's check) should use this validated env without update. Corresponding issue: #366. Will be reworked and suggested again after resolve of it.

PR #346, opened 2023-04-03
Bump version to 0.5.0dev

alexbaden:


kurapov-peter:
I think we already merged it in #344?

PR #347, opened 2023-04-03
[CI] Enable Python style check.

Devjiu:
This PR runs Python(black) style check.


Devjiu:
Based on #345

ienkovich:
@Devjiu Since #345 is closed now, you probably want to make the Python part on top of the existing style check.

@alexbaden Do we still need these *.py files in omniscidb dir?

alexbaden:
> @alexbaden Do we still need these *.py files in omniscidb dir?

Do you mean the files in the Benchmarks dir? Probably not.

Devjiu:
> @Devjiu Since #345 is closed now, you probably want to make the Python part on top of the existing style check.
> 
> @alexbaden Do we still need these *.py files in omniscidb dir?

Resolved, review again, please.

ienkovich (approved):

ienkovich:
Optional change: include black to the env so that everyone uses the same version in their dev env.

.github/workflows/test.yml
```diff
@@ -36,10 +36,19 @@ jobs:
...
+      - name: C++ style check
         if: inputs.test == 'style'
         run: |
           conda run --no-capture-output -n omnisci-dev bash -c "find . -name python -prune -or -name build -prune -or -name ThirdParty -prune -or -name third_party -prune -or \( -name '*.cpp' -print -or -name '*.h' -print \) | xargs clang-format --dry-run --Werror -style=file"
+        shell: bash
+
+      - name: Py style check
+        if: inputs.test == 'style'
+        run: |
+          source $CONDA/etc/profile.d/conda.sh
```
kurapov-peter, line 48: I don't know if this conda activation approach would work in every case. When conda is installed with default settings, the recommended activation script is:
```
eval "$($PATH_TO_CONDA_ROOT/bin/conda shell.$YOUR_SHELL hook)"
```
Not having to escape characters in complex commands is nice though.

PR #348, opened 2023-04-03
Add ResultSetRegistry storage [2/N]

ienkovich:
This PR adds lazy chunk stats to ChunkMetadata.

Currently, we may have lazily computed `ChunkMetadataMap` in `FragmentInfo` when it holds a `ResultSet`. Here I move laziness into `ChunkMetadata` to allow more granular stats computation (i.e. compute stats for some columns only) and use `std::function` instead of `ResultSet` to potentially utilize it later for `ArrowStorage`. Also, only stats are computed lazily now, not the whole `ChunkMetadata`. I'm not sure if we need laziness for `numElements` and `numBytes`. Computing the number of rows for ResultSet can take some time, but it is inevitable anyway so we can do it on import.

### DIFF
--- a/omniscidb/DataMgr/ChunkMetadata.h
+++ b/omniscidb/DataMgr/ChunkMetadata.h
@@ -154,34 +154,67 @@ inline void mergeStats(ChunkStats& lhs,
   }
 }
 
-struct ChunkMetadata {
-  const hdk::ir::Type* type;
-  size_t numBytes;
-  size_t numElements;
-  ChunkStats chunkStats;
+class ChunkMetadata {
+ public:
+  using StatsMaterializeFn = std::function<void(ChunkStats&)>;
+
+  ChunkMetadata(const hdk::ir::Type* type,
+                const size_t num_bytes,
+                const size_t num_elements)
+      : type_(type), num_bytes_(num_bytes), num_elements_(num_elements) {}
+
+  ChunkMetadata(const hdk::ir::Type* type,
+                const size_t num_bytes,
+                const size_t num_elements,
+                const ChunkStats& chunk_stats)
+      : type_(type)
+      , num_bytes_(num_bytes)
+      , num_elements_(num_elements)
+      , chunk_stats_(chunk_stats) {}
+
+  ChunkMetadata(const hdk::ir::Type* type,
+                const size_t num_bytes,
+                const size_t num_elements,
+                StatsMaterializeFn stats_materialize_fn)
+      : type_(type)
+      , num_bytes_(num_bytes)
+      , num_elements_(num_elements)
+      , stats_materialize_fn_(std::move(stats_materialize_fn)) {}
+
+  const hdk::ir::Type* type() const { return type_; }
+  size_t numBytes() const { return num_bytes_; }
+  size_t numElements() const { return num_elements_; }
+  const ChunkStats& chunkStats() const {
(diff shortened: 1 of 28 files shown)


omniscidb/DataMgr/ChunkMetadata.h
```diff
@@ -154,76 +154,134 @@ inline void mergeStats(ChunkStats& lhs,
...
+                                       : type_) &&
+           chunk_stats_.has_nulls == that.chunk_stats_.has_nulls;
   }
+
+ private:
+  void maybeMaterializeStats() const {
+    if (stats_materialize_fn_) {
+      stats_materialize_fn_(chunk_stats_);
+      StatsMaterializeFn().swap(stats_materialize_fn_);
+    }
```
alexbaden, line 277: Should there be some logging or fatal error for failure mode here?

omniscidb/DataProvider/TableFragmentsInfo.h
```diff
@@ -82,7 +80,6 @@ class FragmentInfo {
   mutable size_t numTuples;
   mutable ChunkMetadataMap chunkMetadataMap;
   mutable bool synthesizedNumTuplesIsValid;
```
alexbaden, line 82: We might be able to remove this -- the invalidate calls were used for update/delete, which is all gone.

omniscidb/QueryEngine/InputMetadata.cpp
```diff
@@ -54,23 +52,30 @@ void InputTableInfoCache::clear() {
...
-bool uses_int_meta(const hdk::ir::Type* col_type) {
-  return col_type->isInteger() || col_type->isDecimal() || col_type->isDateTime() ||
-         col_type->isBoolean() || col_type->isExtDictionary();
-}
-
 TableFragmentsInfo synthesize_table_info(hdk::ResultSetTableTokenPtr token) {
   std::vector<FragmentInfo> result;
   bool non_empty = false;
-  for (int frag_id = 0; frag_id < token->resultSetCount(); ++frag_id) {
+  for (int frag_id = 0; frag_id < (int)token->resultSetCount(); ++frag_id) {
```
alexbaden, line 58: I have always preferred `static_cast`, but I wouldn't change it if that's the only change you need to make in this PR.

alexbaden (approved):
A few notes, but overall looks good.

omniscidb/DataMgr/ChunkMetadata.h
ienkovich: I'd expect any logging and exceptions to be in the materializer because that's where synchronization happens. But I don't expect stats computation ever fail because we can always use the default stats.

omniscidb/DataProvider/TableFragmentsInfo.h
ienkovich: You are right, it's not used anymore.

omniscidb/QueryEngine/InputMetadata.cpp
ienkovich: Fixed

omniscidb/DataProvider/TableFragmentsInfo.h
ienkovich: No, I was wrong. It is removed in later patches when we don't store result sets there anymore.

PR #349, opened 2023-04-03
Remove glog dependency

alexbaden:
Folly has a dependency on gflags, not glog. I was able to remove the dependency on glog (we use Boost as the backend for our customer logger impl) by inserting gflags when Folly is enabled.


alexbaden:
The primary motivation for this was to fix the `gflags_shared` linking errors I was seeing w/ Folly on Ubuntu 22.04. I took the opportunity to try and clean up some other items at the same time.

CMakeLists.txt
```diff
@@ -100,7 +100,8 @@ if(ENABLE_FOLLY)
   else()
     set(FOLLY_LIBRARIES "")
     add_definitions("-DHAVE_FOLLY")
-    list(APPEND Folly_LIBRARIES Folly::folly ${Glog_LIBRARIES})
+    find_package(gflags CONFIG REQUIRED)
+    list(APPEND Folly_LIBRARIES Folly::folly)
```
kurapov-peter, line 104: cmake variables are case sensitive, did you mean to add it to the `FOLLY_LIBRARIES`?

PR #360, opened 2023-04-05
Add conda flow for intel gpus

kurapov-peter:


kurapov-peter:
I was able to reproduce it in a clean Ubuntu:22.04 and 20.04 docker. Reproduction steps are here: https://github.com/intel/compute-runtime/issues/595#issuecomment-1497677922

leshikus (commented):
It seems checks fail

leshikus (approved):
LGTM

PR #362, opened 2023-04-05
Unified build flow

kurapov-peter:
A unified build requires both cuda and l0 dependencies installation. For that, a convenient way would be to list all the required platform dependencies as a job parameter. Unfortunately, the language doesn't support lists. Here's a simple workaround.


leshikus (approved):
LGTM

kurapov-peter:
@alexbaden, @ienkovich, we can run this daily instead of on each PR. Thoughts?

PR #365, opened 2023-04-06
Add ResultSetRegistry storage [3/N] 

ienkovich:
With this patch, we finally start using ResultSetRegistry through DataProvider and SchemaProvider interfaces. So, we don't use negative table ids anymore and those code branches can be removed (I removed those I could find but some might remain).

### DIFF
--- a/omniscidb/Calcite/SchemaJson.cpp
+++ b/omniscidb/Calcite/SchemaJson.cpp
@@ -14,6 +14,8 @@
 
 #include "SchemaJson.h"
 
+#include "ResultSetRegistry/ResultSetRegistry.h"
+
 #include <rapidjson/document.h>
 #include <rapidjson/istreamwrapper.h>
 #include <rapidjson/ostreamwrapper.h>
@@ -186,9 +188,20 @@ std::string schema_to_json(SchemaProviderPtr schema_provider) {
   if (dbs.empty()) {
     return "{}";
   }
-  // Current JSON format supports a single database only.
-  CHECK_EQ(dbs.size(), (size_t)1);
-  auto tables = schema_provider->listTables(dbs.front());
+  // Current JSON format supports a single database only. So, we exclude
+  // ResultSetRegistry from the schema for now (which makes it impossible
+  // to run SQL queries on result sets).
+  int db_id;
+  if (dbs.size() == (size_t)1) {
+    db_id = dbs.front();
+  } else {
+    CHECK_EQ(dbs.size(), (size_t)2);
+    CHECK(dbs[0] == hdk::ResultSetRegistry::DB_ID ||
+          dbs[1] == hdk::ResultSetRegistry::DB_ID);
+    db_id = dbs[0] == hdk::ResultSetRegistry::DB_ID ? dbs[1] : dbs[0];
+  }
+
+  auto tables = schema_provider->listTables(db_id);
 
   rapidjson::Document doc(rapidjson::kObjectType);
 
--- a/omniscidb/DataMgr/BufferMgr/Buffer.h
+++ b/omniscidb/DataMgr/BufferMgr/Buffer.h
@@ -137,11 +137,25 @@ class Buffer : public AbstractBuffer {
 
   inline int unPin() override {
     std::lock_guard<std::mutex> pin_lock(pin_mutex_);
-    return (--pin_count_);
+    int res = (--pin_count_);
+    if (!res && delete_on_unpin_) {
+      delete this;
+    }
+    return res;
   }
   inline int getPinCount() override {
     std::lock_guard<std::mutex> pin_lock(pin_mutex_);
     return (pin_count_);
+  }
+
+  inline void deleteWhenUnpinned() override {
+    std::unique_lock<std::mutex> pin_lock(pin_mutex_);
+    if (pin_count_) {
+      delete_on_unpin_ = true;
+    } else {
+      pin_lock.unlock();
+      delete this;
+    }
   }
 
   // Added for testing.
@@ -171,6 +185,7 @@ class Buffer : public AbstractBuffer {
   int epoch_;  /// indicates when the buffer was last flushed
   std::vector<bool> page_dirty_flags_;
   int pin_count_;
+  bool delete_on_unpin_;
   std::mutex pin_mutex_;
   std::unique_ptr<AbstractDataToken> token_;
 };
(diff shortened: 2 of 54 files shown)


ienkovich:
I don't see why ASAN would require more memory than usual here. I tried to put check that the registry is empty at the end of `ArrowSQLRunner::c` method and it worked fine, so all tables are cleaned up from the registry after each query. There might be some buffers cached in BufferMgr and holding some ResultSets alive, but I doubt there can be too many of them. Will try to add some cleanup for such buffers though.

ienkovich:
Looks like ASAN test actually reveals some serious problems. Vagrind shows that we do allocate a lot of memory for the test suite (like ~15x more memory compared to the main branch). If I read the profile correctly, the problem is in the allocation of really huge ResultSets. Probably, there is some issue in the meta info we get for ResultSet tables. Will dig more into this.

ienkovich:
@alexbaden @kurapov-peter 
I've finally resolved the OOM issue for ASAN. The problem was in zero-copy buffers that weren't pinned but still stayed in BufferMgr. These buffers held ResultSets that held RowSetMemoryOwners. These buffers weren't cleared up and lived until the end of the test. I made some dumps and found that those hanging RowSetMemoryOwners allocated only ~100MB in total, but each such owner holds multiple arenas that may have much more actually allocated memory.
To resolve this problem I stopped keeping zero-copy buffers at the BufferMgr indexes and destroy them on unpin. This makes it impossible to re-use the same buffer if we want to fetch a chunk multiple times, but zero-copy fetching is cheap, and removing unneeded ResultSets looks more important now.

omniscidb/Calcite/SchemaJson.cpp
```diff
@@ -186,9 +188,20 @@ std::string schema_to_json(SchemaProviderPtr schema_provider) {
...
+  // ResultSetRegistry from the schema for now (which makes it impossible
+  // to run SQL queries on result sets).
+  int db_id;
+  if (dbs.size() == (size_t)1) {
+    db_id = dbs.front();
+  } else {
+    CHECK_EQ(dbs.size(), (size_t)2);
+    CHECK(dbs[0] == hdk::ResultSetRegistry::DB_ID ||
+          dbs[1] == hdk::ResultSetRegistry::DB_ID);
+    db_id = dbs[0] == hdk::ResultSetRegistry::DB_ID ? dbs[1] : dbs[0];
```
kurapov-peter, line 201: Could you please add a comment on what's going on here? Looks cryptic

omniscidb/QueryEngine/WorkUnitBuilder.cpp
```diff
@@ -327,8 +327,14 @@ void WorkUnitBuilder::process(const ir::Node* node) {
...
         col_var = ir::makeExpr<ir::ColumnVar>(scan->getColumnInfo(i), rte_idx);
       } else {
-        col_var = ir::makeExpr<ir::ColumnVar>(
-            node->getOutputMetainfo()[i].type(), -node->getId(), i, rte_idx, false);
+        CHECK(temporary_tables_.count(-node->getId()));
+        auto& token = temporary_tables_.at(-node->getId());
+        col_var = ir::makeExpr<ir::ColumnVar>(node->getOutputMetainfo()[i].type(),
+                                              token->dbId(),
+                                              token->tableId(),
+                                              i + 1,
```
kurapov-peter, line 335: What is the `+1` for?

kurapov-peter (changes requested):

omniscidb/DataMgr/BufferMgr/Buffer.h
```diff
@@ -137,13 +136,27 @@ class Buffer : public AbstractBuffer {
...
     return (pin_count_);
   }
 
+  inline void deleteWhenUnpinned() override {
+    std::unique_lock<std::mutex> pin_lock(pin_mutex_);
+    if (pin_count_) {
+      delete_on_unpin_ = true;
+    } else {
+      pin_lock.unlock();
+      delete this;
```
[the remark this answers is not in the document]
ienkovich: Actually, it doesn't differ much from the existing situation, when unpinned buffer can be removed by BufferMgr at any time. So, it's not allowed for anyone to use unpinned buffers except BufferMgr, which doesn't store buffers marked as delete-on-unpin.

omniscidb/Calcite/SchemaJson.cpp
ienkovich: There is an explaining comment above saying that we exclude ResultSetRegistry from the schema we serialize to JSON right now.

omniscidb/QueryEngine/WorkUnitBuilder.cpp
ienkovich: In schemas, we always avoided zero ids for dbs/tables/cols/frags, so column_id for ResultSet column is its index + 1.

omniscidb/Calcite/SchemaJson.cpp
kurapov-peter: The confusing part to me is simply that the comment reads "There's only one database" and then there're two in the `dbs[0]`, `dbs[1]`.

omniscidb/QueryEngine/WorkUnitBuilder.cpp
kurapov-peter: Sounds error-prone :)

omniscidb/Calcite/SchemaJson.cpp
ienkovich: Sorry, I don't see these words in my comment :) The comment says: "Current JSON format supports a single database only. So, we exclude ResultSetRegistry from the schema for now". We do it by choosing db_id which is not the registry.

I can rephrase my comment if it is not clear enough. But this part is also modified by later patches in the series to support SQL over result sets.

kurapov-peter: Alright

kurapov-peter (approved):

omniscidb/QueryEngine/WorkUnitBuilder.cpp
ienkovich: Hmm... I can add a method to ResultSetToken to get column id by column index. Such incapsulation would make code more clear. I'm not sure this would cover all `+1` cases though.

PR #368, opened 2023-04-06
Add an issue bucketing script for query execution tests

kurapov-peter:
This is a basic version that collects test statuses in presence of failures. It captures the return codes of standalone tests and creates a yaml report. The report is coarse-grained since each test may include many queries, but it gives some coverage picture. The intended use is to simplify status checking for non-standard build/run options (like `exec.heterogeneous.allow_cpu_retry = false`). For example, I've collected the stats for `ArrowBasedExecuteTest` on Intel GPU (we have 211 tests in total there, each includes multiple queries):

| status | number of tests |
| --- | --- |
| passes | 116 |
| assertion/driver failure | 82 |
| incorrect result | 13 |

Bucketing granularity can be easily improved by capturing the stderr. The help should be self-explanatory. Let me know if anything is unclear.

### DIFF
--- a/scripts/bucketing.py
+++ b/scripts/bucketing.py
@@ -1,0 +1,84 @@
+#!/usr/bin/python3
+
+import subprocess
+import argparse
+import pathlib
+from string import whitespace
+import yaml
+import signal
+
+
+def signal_to_name(sig: int):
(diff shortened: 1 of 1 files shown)


leshikus (approved):
LGTM

PR #369, opened 2023-04-06
Fix docker cache

leshikus:
Subj


leshikus:
Image build time decreased from 42m to 3m

alexbaden:
What Docker image repository are we using?

.github/workflows/build-docker.yml
```diff
@@ -11,6 +11,9 @@ on:
       options:
         type: string
         default: -DENABLE_CUDA=off
+    secrets:
+      DOCKERHUB_PASSWORD:
```
Devjiu, line 15: I thought that secrets are global vars for all github jobs, but after discussion with Alexei I understood that it's special github behavior https://github.blog/changelog/2022-05-03-github-actions-simplify-using-secrets-with-reusable-workflows/ . Maybe there are some WA, because secrets added as https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-an-environment . 

This is only to mark and summarize our discussion.

Devjiu (approved):
👍 Great PR. 
Currently it uses https://hub.docker.com/u/dataved as public image and pushed new image on combined Dockerfile hash changed.  As a result dockerhub is our public cache. 

Docker pull step takes about ~3 minutes, that is better than docker build even using cached layers. 

Please, provide description to this pr, because there are a lot of fundamental changes in Docker build workflow (files moved from omnisci folder and now used public image). 

Also currently Docker image update happens automatically (can happen on each pr) I would prefer a separate job to handle changes to the image, but this is not critical, we will test this suggested approach and update later. 

There are also some minor motes, that can be done in further prs.

PR #383, opened 2023-04-13
Add the LLVM MulOverFlow, extend IntelGPUtest

lmontigny:


omniscidb/QueryEngine/ArithmeticIR.cpp
```diff
@@ -385,29 +385,61 @@ llvm::Value* CodeGenerator::codegenMul(const hdk::ir::BinOper* bin_oper,
...
+    auto cmp_plus = cgen_state_->ir_builder_.CreateICmpUGT(lhs_pos, div_limit_plus);
+    auto neg_overflow = cgen_state_->ir_builder_.CreateAnd(args_is_neg, cmp_plus);
+    cgen_state_->ir_builder_.CreateCondBr(neg_overflow, mul_fail, mul_check_2);
+    cgen_state_->ir_builder_.SetInsertPoint(mul_check_2);
+
+    // if !(IsNegative) && UX > (static_cast<U>(std::numeric_limits<T>::max())) / UY;
+    auto div_limit = cgen_state_->ir_builder_.CreateUDiv(chosen_max, rhs_pos);
+    auto cmp = cgen_state_->ir_builder_.CreateICmpUGT(lhs_pos, div_limit);
+    auto pos_overflow = cgen_state_->ir_builder_.CreateAnd(args_is_pos, cmp);
+    cgen_state_->ir_builder_.CreateCondBr(pos_overflow, mul_fail, mul_ok);
```
kurapov-peter, line 442: I think there's still a problem with the implementation. The `arg_is_pos`/`arg_is_neg` should have been calculated as `xor` not `or` (as they are in the link that you've shared btw).
Consider the following simplified example:
Let's have a bit width of 3:
| bit representation | signed value | unsigned value |
| --- | --- | --- |
| 000 | 0 | 0 |
| 001 | 1 | 1 |
| 010 | 2 | 2 |
| 011 | 3 | 3 |
| 100 | -4 | 4 |
| 101 | -3 | 5 |
| 110 | -2 | 6 |
| 111 | -1 | 7 |

Consider a * b: a = -2, b = -2. This is a positive overflow case. Then:
* `lhs_is_neg = rhs_is_neg = true`, thus `args_is_neg = true`
* `lhs_is_pos = rhs_is_pos = false`, thus `args_is_pos = false`
* In our case signed max is `3`, so `max_plus_one = 4`, and `div_limit_plus = 4/2 = 2`
* `div_lim = 3/2 = 1`
* `cmp_plus = 2 > 2 = false`, so no negative overflow, but
* `cmp = 2 > 1 = true`, `pos_overflow = false and true = false`, which is incorrect.

I also suspect you don't need the pair of positive/negative cases since you already got rid of the zeroes case in the first basic block. Xoring the arguments signedness will determine the case to check.

kurapov-peter (changes requested):

omniscidb/Tests/IntelGPUEnablingTest.cpp
```diff
@@ -445,6 +445,26 @@ TEST_F(AggregationTest, NullHandling) {
...
+class OverflowTest : public ExecuteTestBase, public ::testing::Test {};
+
+TEST_F(OverflowTest, OverflowAndUnderFlow) {
+  c("SELECT dd * ufd FROM test;", g_dt);
+  c("SELECT dd * -1 FROM test;", g_dt);
+  c("SELECT -1 * dd FROM test;", g_dt);
+  c("SELECT ofq * -1 FROM test;", g_dt);
+  c("SELECT 1 * ofq FROM test;", g_dt);
+  c("SELECT 566 * 244 FROM test;", g_dt);
+  EXPECT_THROW(c("SELECT ofd * ofd FROM test;", g_dt), std::runtime_error);
```
kurapov-peter, line 457: This can be a more reliable test if you don't run the comparison. The `runtime_error` may come from sqlite or the comparison itself, so we'll not catch an issue with hdk not detecting an overflow.

kurapov-peter (changes requested):

kurapov-peter (approved):
Looks good to me now!

Before we merge, let's make sure the CI is green and could you please remove the `from` clause from queries like `SELECT 9223372036854775807 * 2 FROM test;`? - looks a bit confusing.

Issue #379, opened 2023-04-12
Execution fails on memory level assertion


kurapov-peter:
This is likely related to strings support that would need to be disabled for now for L0.
Failing test list:
 - Select.FilterShortCircuit
  - Select.InValues
  - Select.ComplexQueries
  - Select.Case
  - Select.Strings                                                                                                   
  - Select.SharedDictionary
  - Select.StringCompare
  - Select.DictionaryStringEquality

kurapov-peter:
I think the Select.Case is the only one left in the current main. @akroviakov, could you please double-check and remove it from the excluded enabling tests?

PR #388, opened 2023-04-14
Temporary disable string queries for L0

akroviakov:
Addresses #379.
Detect string related queries and fall back to CPU from L0, extend memory level handling for L0.

- Select.FilterShortCircuit - runs on L0
- Select.InValues - runs on L0
- Select.ComplexQueries - fallback to CPU (on string-related queries)
- Select.Case - fallback to CPU (on string-related queries) / `PerfectJoinHashTable` is not functional for L0
- Select.Strings - fallback to CPU
- Select.SharedDictionary - fallback to CPU
- Select.StringCompare - fallback to CPU
- Select.DictionaryStringEquality - explicitly disallows CPU retry

### DIFF
--- a/omniscidb/QueryEngine/WorkUnitBuilder.cpp
+++ b/omniscidb/QueryEngine/WorkUnitBuilder.cpp
@@ -440,13 +440,14 @@ void WorkUnitBuilder::processFilter(const ir::Filter* filter) {
       executor_, input_nest_levels_, join_types_, now_, eo_.just_explain);
   auto rte_idx = all_nest_levels_.at(filter);
 
-#ifdef HAVE_L0
-  StringGuardForL0 strConstCollector;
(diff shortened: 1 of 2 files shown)


omniscidb/QueryEngine/WorkUnitBuilder.cpp
```diff
@@ -423,6 +440,14 @@ void WorkUnitBuilder::processFilter(const ir::Filter* filter) {
       executor_, input_nest_levels_, join_types_, now_, eo_.just_explain);
   auto rte_idx = all_nest_levels_.at(filter);
 
+#ifdef HAVE_L0
```
kurapov-peter, line 443: Can we query the platform type instead of the macro here?

kurapov-peter (changes requested):

kurapov-peter (approved):
Thanks!

PR #389, opened 2023-04-15
Linear regression backup

bagrorg:
This is the #326, but with fixed history


omniscidb/QueryEngine/CostModel/DataSources/DwarfBench.cpp
```diff
@@ -13,69 +13,41 @@
...
-    : DataSource(DataSourceConfig{.dataSourceName = "DwarfBench",
-                                  .supportedDevices = {ExecutorDeviceType::CPU},
-                                  .supportedTemplates = {AnalyticalTemplate::GroupBy,
-                                                         AnalyticalTemplate::Join,
-                                                         AnalyticalTemplate::Scan}}) {}
+struct DwarfBenchDataSource::PrivateImpl {
+  DwarfBench::Dwarf convertToDwarf(AnalyticalTemplate templ);
+  DwarfBench::DeviceType convertDeviceType(ExecutorDeviceType device);
+  std::vector<Detail::Measurement> convertMeasurement(
+      const std::vector<DwarfBench::Measurement> measurements);
```
kurapov-peter, line 29: What's the reason you're passing the vector by value?

omniscidb/QueryEngine/Execute.cpp
```diff
@@ -1831,8 +1835,25 @@ hdk::ResultSetTable Executor::executeWorkUnitImpl(
     exe_policy = std::make_unique<policy::FragmentIDAssignmentExecutionPolicy>(
         ExecutorDeviceType::CPU);
   } else {
+#ifdef HAVE_COST_MODEL
     costmodel::DummyCostModel cost_model(device_type, config_->exec.heterogeneous);
```
kurapov-peter, line 1839: The idea behind having a dummy cost model was to not pollute the code with macros. Can we have the model library handle it internally and return a round-robin policy (or whichever else requested)?

kurapov-peter (changes requested):

omniscidb/QueryEngine/Execute.cpp
bagrorg: that's exactly what we're doing in `predict`. define is here to handle the case when we don't use a library with a cost model. alternatively, we can assume that we always use it and remove the ability to turn it off in cmake.
or maybe I misunderstood the comment?

omniscidb/QueryEngine/CostModel/DataSources/DwarfBench.cpp
bagrorg: oops =)

kurapov-peter (approved):

PR #394, opened 2023-04-17
Fixed inline functions declaration on windows

gshimansky:
If inline flag is present on windows then bodies of inline functions is not generated (they all are inlined) and then JIT fails to find these symbols.


gshimansky:
Closed because CI doesn't work from a forked repo.

PR #396, opened 2023-04-17
Fixed generation of left shift on windows

gshimansky:
`__lzcnt` counts leading zeroes in a 32-bit value, different from what `__builtin_ffs` does on Linux.


gshimansky:
Closed because CI doesn't work from a forked repo.

PR #403, opened 2023-04-17
Updated conda env for windows development

gshimansky:


alexbaden (approved):
Looks good. We should look at merging these changes in w/ the Linux version. I don't know if there's a way to do that inside conda, but we could write a shell script to concat them.

gshimansky:
We could probably have some common environment for both platforms and small additions that exist/required on specific platforms (e.g. folly currently exists only on Linux in conda-forge).

gshimansky:
I suggest that we murge this PR while we think on the best way to synchronize multi-platform conda env. This is the last PR that is holding windows build from running CI.

aregm:
> I suggest that we merge this PR while we think on the best way to synchronize multi-platform conda env. This is the last PR that is holding windows build from running CI.

I think we should merge this ASAP, as this is a separate file and do not interfere with Linux builds.

aregm (approved):

PR #406, opened 2023-04-17
Fixed inline functions declaration on windows

gshimansky:
If inline flag is present on windows then bodies of inline functions is not generated and then JIT fails to find these symbols.


alexbaden (approved):
Not entirely clear it does anything on Linux, either.

PR #407, opened 2023-04-17
Fixed call to gmtime_s because error code is ignored on Linux

gshimansky:
The `nsec` value passed to `gmtime_s` (on Windows) and `gmtime_r` (on Linux) is the same but on Linux we ignore an error code and test passes.


alexbaden (approved):
Do we need the logging?

gshimansky:
> Do we need the logging?

I am not sure. I was just thinking that maybe erroneous value that causes gmtime to return an error should not be there, so we could look into it later and fix the test.

PR #417, opened 2023-04-19
324 replace the llvm legacy pass manager with the llvm new pass manager clean

lmontigny:


kurapov-peter (changes requested):
So, apart from the minor comment I think the PR is good. I propose following steps to make sure we don't have a perf regression:
1. Check the functional testing time does not deviate significantly from the main.
2. Run benchmarks on small amount of data (or time the compilation specifically).

lmontigny:
Functional testing timing:
[with new PassManager](https://gist.github.com/lmontigny/907e22bbb14bc88ed4417b39714af25b)
[with legacy PassManager](https://gist.github.com/lmontigny/e2c2e5f65d827de9690984ea18d119b4)

kurapov-peter:
> Functional testing timing: [with new PassManager](https://gist.github.com/lmontigny/907e22bbb14bc88ed4417b39714af25b) [with legacy PassManager](https://gist.github.com/lmontigny/e2c2e5f65d827de9690984ea18d119b4)

Would be even more important to check how does CPU perform (cuda as well).

lmontigny:
- Compare full sanity check from CI directly after rebase
- Compare timing using omniscript benchmarks

kurapov-peter (approved):

PR #423, opened 2023-04-21
Cost model integration

bagrorg:
This PR introduces the use of the `cost model` during query execution


omniscidb/QueryEngine/CostModel/IterativeCostModel.cpp
```diff
@@ -0,0 +1,40 @@
...
+        size_t gpuSize = queryInfo.bytesSize - curSize;
+
+        size_t cpuPrediction = dp_[ExecutorDeviceType::CPU][queryInfo.templ]->getExtrapolatedData(cpuSize);
+        size_t gpuPrediction = dp_[ExecutorDeviceType::GPU][queryInfo.templ]->getExtrapolatedData(gpuSize);
+        size_t curPrediction = std::max(gpuPrediction, cpuPrediction);
+
+        if (curPrediction < runtimePrediction) {
+            runtimePrediction = curPrediction;
+
+            cpuProp = static_cast<float>(cpuSize) / queryInfo.bytesSize * 10;
```
kurapov-peter, line 27: What does the number `10` mean?

omniscidb/QueryEngine/CostModel/IterativeCostModel.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <cmath>
+
+namespace costmodel {
+
+std::unique_ptr<policy::ExecutionPolicy> IterativeCostModel::predict(
+    QueryInfo queryInfo) {
+        std::lock_guard<std::mutex> l(latch_);
+
+    unsigned cpuProp, gpuProp;
+    size_t optStep = std::ceil(static_cast<float>(queryInfo.bytesSize) / optimizationIterations);
```
kurapov-peter, line 13: Please stick to the style guide: https://google.github.io/styleguide/cppguide.html#Variable_Names

omniscidb/QueryEngine/CostModel/CostModel.cpp
```diff
@@ -49,15 +53,21 @@ void CostModel::calibrate(const CaibrationConfig& conf) {
 
     for (auto& templateMeasurement : dmEntry.second) {
       AnalyticalTemplate templ = templateMeasurement.first;
+
+#ifdef HAVE_ARMADILLO
+      dp_[device][templ] =
+          std::make_unique<LinearRegression>(std::move(templateMeasurement.second));
+#else
```
kurapov-peter, line 60: Looks like this can be resolved by inheritance instead of macros.

omniscidb/QueryEngine/CostModel/IterativeCostModel.cpp
```diff
@@ -0,0 +1,40 @@
+#include "IterativeCostModel.h"
+#include "Dispatchers/ProportionBasedExecutionPolicy.h"
+
+#include <cmath>
+
+namespace costmodel {
+
+std::unique_ptr<policy::ExecutionPolicy> IterativeCostModel::predict(
+    QueryInfo queryInfo) {
+        std::lock_guard<std::mutex> l(latch_);
```
kurapov-peter, line 10: I don't mind having a non-const `predict` method, in general, to account for a feedback loop in the future (even though it might be cleaner to have a separate update method), but the locking here looks way too pessimistic. Do we really need to synchronize here? If a critical section is unavoidable, can we at least narrow the scope?

omniscidb/QueryEngine/Execute.cpp
```diff
@@ -1796,6 +1813,79 @@ hdk::ResultSetTable Executor::finishStreamExecution(
   return result;
 }
 
+std::pair<std::unique_ptr<policy::ExecutionPolicy>, ExecutorDeviceType>
```
kurapov-peter, line 1816: Don't we have the execution device types captured by the `ExecutionPolicy`?

omniscidb/QueryEngine/Execute.cpp
```diff
@@ -1796,6 +1813,79 @@ hdk::ResultSetTable Executor::finishStreamExecution(
...
+
+    costmodel::QueryInfo qi = {.templ = ra_exe_unit.templ, .bytesSize = bytes};
+
+    // TODO check that template is available for cost model
+    exe_policy = ra_exe_unit.cost_model->predict(qi);
+    return {std::move(exe_policy), requested_device_type};
+  }
+
+  LOG(DEBUG1) << "Cost Model disabled, or template is unknown: templ="
+              << costmodel::templateToString(ra_exe_unit.templ)
```
kurapov-peter, line 1867: There's a common mechanism for all the to-string methods in HDK, please have a look at https://github.com/intel-ai/hdk/blob/main/omniscidb/Shared/toString.h.

omniscidb/QueryEngine/Execute.cpp
```diff
@@ -2120,6 +2196,13 @@ void Executor::addTransientStringLiterals(
 ExecutorDeviceType Executor::getDeviceTypeForTargets(
     const RelAlgExecutionUnit& ra_exe_unit,
     const ExecutorDeviceType requested_device_type) {
+  if (isDevicesRestricted(ra_exe_unit, requested_device_type))
+    return ExecutorDeviceType::CPU;
+  return requested_device_type;
+}
+
+bool Executor::isDevicesRestricted(const RelAlgExecutionUnit& ra_exe_unit,
```
kurapov-peter, line 2204: Please consider renaming the method to smth more descriptive.

omniscidb/QueryEngine/ExecutionKernel.cpp
```diff
@@ -150,6 +150,10 @@ void ExecutionKernel::run(Executor* executor,
   }
 }
 
+ExecutorDeviceType ExecutionKernel::getChosenDevice() const {
```
kurapov-peter, line 153: Probably better `getChosenDeviceType`. Anyway, I think it's better to provide a `toString` method for all the logging.

omniscidb/QueryEngine/RelAlgExecutionUnit.h
```diff
@@ -30,6 +30,8 @@
 #include "RelAlgDagBuilder.h"
 #include "Shared/sqldefs.h"
 #include "Shared/toString.h"
+#include "CostModel/Measurements.h"
+#include "CostModel/CostModel.h"
```
kurapov-peter, line 34: Can we get by a single header?

omniscidb/Shared/Config.h
```diff
@@ -107,6 +107,8 @@ struct ExecutionConfig {
 
   bool materialize_inner_join_tables = true;
   std::string initialize_with_gpu_vendor = "";
+
+  bool use_cost_model = false;
```
kurapov-peter, line 111: I think `enable_cost_model` would be more consistent, wouldn't it?

omniscidb/QueryEngine/Execute.cpp
```diff
@@ -1796,6 +1813,79 @@ hdk::ResultSetTable Executor::finishStreamExecution(
...
+                ExecutorDeviceType::CPU),
+            ExecutorDeviceType::CPU};
+  }
+
+  CHECK(!query_infos.empty());
+  if (!max_groups_buffer_entry_guess) {
+    LOG(DEBUG1) << "The query has failed the first execution attempt because of running "
+                   "out of group by slots. Make the conservative choice: allocate "
+                   "fragment size slots and run on the CPU.";
+    max_groups_buffer_entry_guess = compute_buffer_entry_guess(query_infos);
```
kurapov-peter, line 1833: Did we move the method into the cost model? Would it create a circular dependency again?

kurapov-peter (changes requested):

bagrorg:
@kurapov-peter, thanks! Could you please also advice something [here](https://github.com/intel-ai/hdk/pull/423/files#diff-fbe12308d9fd5fec3b3e1ad5405f2e262892386ed5e8702beccedbe3be3f65f8R1850)? I'm not sure if this is a good solution

omniscidb/QueryEngine/Execute.cpp
bagrorg: We put only the end of this method in the `cost model`, but it seemed to me that it was worth bringing it here in case the `cost model` is disabled. the `cost model` itself does not use this code, so there should be no cyclic dependency. or did I misunderstand the comment?

omniscidb/QueryEngine/Execute.cpp
bagrorg: previously, the device type from `CompilationOptions` was used in `collectAllDeviceResults` and for non-heterogeneous execution. I left it to preserve what has already been written. It seemed to me that it was not necessary to store this type, because, indeed, everything is in `exe_policy`. but in order not to accidentally break anything, it seemed to me that it was better to leave the logic that is already written in `executeWorkUnitImpl`, so as not to mix refactoring of this section and the integration of the cost model in one pull request

omniscidb/QueryEngine/CostModel/CostModel.cpp
bagrorg: I made a provider for extrapolation, but still left defines. just took them out into it. maybe I misunderstood this comment anyway?

omniscidb/QueryEngine/CostModel/CostModel.h
```diff
@@ -14,37 +14,56 @@
...
 struct CaibrationConfig {
   std::vector<ExecutorDeviceType> devices;
 };
 
+struct QueryInfo {
+  std::vector<AnalyticalTemplate> templs;
+  size_t bytes_size;
+};
+
+struct CostModelConfig {
```
kurapov-peter, line 37: Do we expect more fields here? I.e. why use a structure for a single entry?

omniscidb/QueryEngine/CostModel/DataSources/DwarfBench.cpp
```diff
@@ -72,12 +72,12 @@ std::vector<Detail::Measurement> DwarfBenchDataSource::PrivateImpl::convertMeasu
...
     : DataSource(DataSourceConfig{
-          .dataSourceName = "DwarfBench",
-          .supportedDevices = {ExecutorDeviceType::CPU, ExecutorDeviceType::GPU},
-          .supportedTemplates = {AnalyticalTemplate::GroupBy,
-                                 AnalyticalTemplate::Join,
-                                 AnalyticalTemplate::Scan,
-                                 AnalyticalTemplate::Sort}})
+          .data_source_name = "DwarfBench",
+          .supported_devices = {ExecutorDeviceType::CPU, ExecutorDeviceType::GPU},
+          .supported_templates = {AnalyticalTemplate::GroupBy,
```
kurapov-peter, line 77: There's an issue with this syntax on windows, unfortunately. Could you please use the old-school plain init lists?

omniscidb/QueryEngine/CostModel/ExtrapolationModels/ExtrapolationModelProvider.cpp
```diff
@@ -0,0 +1,33 @@
...
+}
+std::unique_ptr<ExtrapolationModel> ExtrapolationModelProvider::provide(
+    std::vector<Detail::Measurement>&& measurement) {
+#ifdef HAVE_ARMADILLO
+  return std::make_unique<LinearRegression>(std::move(measurement));
+#else
+  return std::make_unique<LinearExtrapolation>(std::move(measurement));
+#endif
+}
+}  // namespace costmodel
```
kurapov-peter, line 33: Missing EOL at EOF

omniscidb/QueryEngine/CostModel/IterativeCostModel.cpp
```diff
@@ -0,0 +1,66 @@
...
+#else
+IterativeCostModel::IterativeCostModel()
+    : CostModel({std::make_unique<EmptyDataSource>()}) {}
+#endif
+
+std::unique_ptr<policy::ExecutionPolicy> IterativeCostModel::predict(
+    QueryInfo query_info) const {
+  std::shared_lock<std::shared_mutex> l(latch_);
+
+  unsigned cpu_prop, gpu_prop;
```
kurapov-peter, line 24: I see only one assignment below under an `if` statement - potentially UB.

omniscidb/QueryEngine/CostModel/IterativeCostModel.h
```diff
@@ -0,0 +1,33 @@
+/*
+    Copyright (c) 2022 Intel Corporation
```
kurapov-peter, line 2: It's 2023 now ;)

omniscidb/QueryEngine/Visitors/TemplateAggregationVisitor.h
```diff
@@ -0,0 +1,94 @@
...
+
+class TemplateAggregationVisitor : public RelAlgVisitor<TemplateSample> {
+ public:
+  virtual TemplateSample visitAggregate(const hdk::ir::Aggregate* n) const {
+    bool consider = true;
+    if (visited_nodes_.find(n) != visited_nodes_.end())
+      consider = false;
+    else
+      visited_nodes_.insert(n);
+    return {consider, costmodel::AnalyticalTemplate::GroupBy};
```
kurapov-peter, line 22:
```suggestion
    bool consider = visited_nodes_.count(n);
    if (!consider) {
      visited_nodes_.insert(n);
    }
    return {consider, costmodel::AnalyticalTemplate::GroupBy};
```
And maybe move the code to some private method to avoid duplication.

omniscidb/QueryEngine/Visitors/TemplateAggregationVisitor.h
```diff
@@ -0,0 +1,94 @@
...
+    collected_templates_.clear();
+
+    return ret;
+  }
+
+ protected:
+  virtual TemplateSample aggregateResult(const TemplateSample&,
+                                         const TemplateSample& next_result) const {
+    if (next_result.consider)
+      collected_templates_.push_back(next_result.templ);
```
kurapov-peter, line 84:
```suggestion
    if (next_result.consider) {
      collected_templates_.push_back(next_result.templ);
    }
```

kurapov-peter (changes requested):

omniscidb/QueryEngine/CostModel/CostModel.h
bagrorg: Yes, I would like to leave it like this, because it seems to me that new arguments may quietly appear in the future.

omniscidb/QueryEngine/CostModel/DataSources/DwarfBench.cpp
bagrorg: oh, hdk exists on windows.. =) i missed it

omniscidb/QueryEngine/CostModel/CostModel.cpp
```diff
@@ -14,50 +14,77 @@
...
+    if (!config_.data_source->isDeviceSupported(device))
       throw CostModelException("device " + deviceToString(device) + " not supported in " +
-                               dataSource_->getName() + " data source");
+                               config_.data_source->getName() + " data source");
   }
 }
 
 void CostModel::calibrate(const CaibrationConfig& conf) {
-  std::lock_guard<std::mutex> g{latch_};
+  std::unique_lock<std::shared_mutex> l(latch_);
```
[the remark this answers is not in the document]
bagrorg: we don't want to be able to recalibrate the cost model? or is it irrelevant without client-server architecture?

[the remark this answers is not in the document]
kurapov-peter: Interactive scenarios don't seem to need recalibration, but it's nice to have the ability to do that. So let's leave it be for now.

kurapov-peter (approved):
LGTM

PR #425, opened 2023-04-21
check maven cache in docker

Devjiu:


Devjiu:
In any case it should be done from scratch. Postponed.

Issue #317, opened 2023-03-23
Date values are incorrectly casted to int


ienkovich:
Right now, we don't allow date values to be cast to integers in the QueryBuilder. It looks like this cast ignores the actual time unit used for the date, but we expect it to produce an integer value according to its time unit (BTW timestamp behavior is different and matches expectations).

PR #427, opened 2023-04-24
[Date Conversion] Enables conversion date to int

Devjiu:
This commit enables conversion from date to int.

Resolves: #317

### DIFF
--- a/omniscidb/QueryBuilder/QueryBuilder.cpp
+++ b/omniscidb/QueryBuilder/QueryBuilder.cpp
@@ -828,7 +828,7 @@ BuilderExpr BuilderExpr::cast(const Type* new_type) const {
       return {builder_, expr_->cast(new_type), "", true};
     }
   } else if (expr_->type()->isDate()) {
-    if (new_type->isDate() || new_type->isTimestamp()) {
+    if (new_type->isNumber() || new_type->isDate() || new_type->isTimestamp()) {
       return {builder_, expr_->cast(new_type), "", true};
     }
   } else if (expr_->type()->isTime()) {
--- a/omniscidb/Tests/QueryBuilderTest.cpp
+++ b/omniscidb/Tests/QueryBuilderTest.cpp
@@ -321,6 +321,21 @@ class QueryBuilderTest : public TestSuite {
                     {"col_vc_10", ctx().varChar(10)},
                 });
 
+    createTable("test_date",
+                {
+                    {"col_bi", ctx().int64()},
+                    {"col_i", ctx().int32()},
+                    {"col_f", ctx().fp32()},
+                    {"col_d", ctx().fp64()},
+                    {"col_dec", ctx().decimal64(10, 2)},
+                    {"col_b", ctx().boolean()},
+                    {"col_str", ctx().text()},
+                    {"col_date", ctx().date32(hdk::ir::TimeUnit::kDay)},
+                });
+    insertCsvValues("test_date",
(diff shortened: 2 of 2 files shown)


omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -828,7 +828,7 @@ BuilderExpr BuilderExpr::cast(const Type* new_type) const {
       return {builder_, expr_->cast(new_type), "", true};
     }
   } else if (expr_->type()->isDate()) {
-    if (new_type->isDate() || new_type->isTimestamp()) {
+    if (new_type->isNumber() || new_type->isDate() || new_type->isTimestamp()) {
```
ienkovich, line 831: `isNumber` includes decimals and FP types. Is it intentional?

omniscidb/Tests/QueryBuilderTest.cpp
```diff
@@ -493,6 +509,28 @@ TEST_F(QueryBuilderTest, Arithmetics) {
...
+TEST_F(QueryBuilderTest, DateToInt) {
+  QueryBuilder builder(ctx(), schema_mgr_, configPtr());
+
+  auto tinfo_a = builder.scan("test_date");
+
+  auto dag = tinfo_a.proj(tinfo_a.ref("col_date").cast("int32")).finalize();
+  auto res = runQuery(std::move(dag));
+  LOG(ERROR) << "res: " << res.toString();
+  LOG(ERROR) << "res: " << toArrow(res)->ToString();
+  compare_res_data(res, std::vector<int32_t>({946684800, 1426636800}));
```
ienkovich, line 521: Are these numbers what you would expect for used dates in `kDay` time unit?

omniscidb/Tests/QueryBuilderTest.cpp
```diff
@@ -321,6 +321,21 @@ class QueryBuilderTest : public TestSuite {
...
+    createTable("test_date",
+                {
+                    {"col_bi", ctx().int64()},
+                    {"col_i", ctx().int32()},
+                    {"col_f", ctx().fp32()},
+                    {"col_d", ctx().fp64()},
+                    {"col_dec", ctx().decimal64(10, 2)},
+                    {"col_b", ctx().boolean()},
+                    {"col_str", ctx().text()},
+                    {"col_date", ctx().date32(hdk::ir::TimeUnit::kDay)},
```
ienkovich, line 333: We need to check other date sizes and time units as well.

omniscidb/QueryBuilder/QueryBuilder.cpp
Devjiu: I took conversion `timestamp` as an example, it uses 'timestamp' -> 'number, date, timestamp'. Maybe we should create some conversion map?

omniscidb/Tests/QueryBuilderTest.cpp
Devjiu: I converted `2000-01-01, 2015-03-18` to int and used this values. (also checked what happens with online converter https://timestamp.online) In general it can be expanded to Hours and Milli.

ienkovich: This converter gives you a timestamp in seconds. I would expect different int values when casting dates in different units.

Devjiu:
For some reason I can't  update pr, so recreating it: #472

PR #442, opened 2023-04-28
Add ResultSetRegistry storage [6/N]

ienkovich:
We now allow ResultSet to be used as an input table but it still lacks support for fetching strings and arrays. This patch adds varlen data fetch to ResultSetRegistry. Fixed-length arrays will be supported separately.


alexbaden (approved):
Looks good and resolves the issue we noticed with multi step queries in the tests.

Issue #447, opened 2023-05-02
[CI] Unify all buidl.yml 


Devjiu:
We have 3 build.yml files: 
- build.yml
- windows-build.yml
- build-docker.yml

Naming is inconsistent and any changes in common build pipeline requires changes in 3 files, what complicates maintenance. 

Also it uses different scripts:
- omniscidb\scripts\conda\build.bat
- omniscidb/scripts/conda/build.sh

That simply wraps common: `cmake -B build -S . $@ && cmake --build build --parallel $(nproc)`

alexbaden:
I don't think the build scripts will change that much, so unifying the shell and batch scripts seems more trouble than it's worth (unless there's something easy I am missing). 

As far as separate build files, I recognize there is duplication, but there are also benefits to keeping the number of steps in the build clean. It can be confusing when you see steps about tests which are skipped because they are not relevant to the current job.

Devjiu:
> I don't think the build scripts will change that much, so unifying the shell and batch scripts seems more trouble than it's worth (unless there's something easy I am missing).
> 
> As far as separate build files, I recognize there is duplication, but there are also benefits to keeping the number of steps in the build clean. It can be confusing when you see steps about tests which are skipped because they are not relevant to the current job.

Main issue here is that build.sh/build.bat not used by developers, so it increases probability of `works on my machine, fails in CI`. I mean that `cmake -B build -S . $@ && cmake --build build --parallel $(nproc)` can be called directly in yml files.

There are 2 possible ways:
1. add compile defs inside of common build.py to unify dev workflow with CI workflow
2. use cmake build explicitly in CI workflow. 

Currently we are using some mix of it and build.sh/build.bat are useless for developers and to reproduce you have to unwrap all those wrappers.

Devjiu:
Also we already have steps, that are skipped during build. 
For example:
![image](https://user-images.githubusercontent.com/14844002/235926899-12db2403-bd15-43fc-9699-be3a8cc2c35f.png)

alexbaden:
Works on my machine fails in CI is very unlikely with simple cmake commands. I don't think the benefit is worth the tradeoff. The script does other things too -- sets up environment variables, etc. 

As far as steps that are skipped during build, that is my point. They can be confusing, I don't want to add more :)

PR #448, opened 2023-05-02
[CI] Naming

Devjiu:
This commit unifies naming after addition of Windows build.

See also: #447

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   build:
-    name: Docker - ${{ inputs.name }}
+    name: Linux - Docker - ${{ inputs.name }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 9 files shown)


kurapov-peter (approved):
Should we explicitly add options to CPU builds like `-DENABLE_CUDA=off -DENABLE_L0=off` to prevent future issues like we have with CUDA build?

Devjiu:
> Should we explicitly add options to CPU builds like `-DENABLE_CUDA=off -DENABLE_L0=off` to prevent future issues like we have with CUDA build?

I think build parameters should be associated with configuration in some map-like structure. Lnx + Conda + cpu = extra_flags/extra options. From the other side it can be done as hierarchically concatenated options.

gshimansky (approved):

alexbaden:
I'm ok with it, but I do think we can spare the extra two letters to make `Linux` more readable. :)

PR #450, opened 2023-05-02
Remove const from input nodes.

ienkovich:
I found it really inconvenient to have `const` qualifier for `Node` inputs. Thus, having a root node we cannot walk and transform the DAG and have to use non-const pointers list stored in the DAG (which cannot even be rebuilt through the DAG walk from the root). I don't see why we should have this `const ` qualifier, `Node` objects are not immutable.


alexbaden (approved):
I don't see a need for `const` either.

PR #453, opened 2023-05-03
[CI] Enable Conda cache

Devjiu:
This commit enables caching check and unifies cache keys.


Devjiu:
Reduces time for conda builds for 5 mins.

Devjiu:
Created several caches. 
Based on:
1. hash + job date
2. runner os: Linux/Windows
3. input.param - cpu/cuda/l0/all-gpus

So currently it will use 5 caches:
1. Windows-cpu
2. Linux cpu
3. Linux cuda
4. Linux l0
5. Linux all-gpus 

Amount if needed can be reduced with sharing base env and installation on steps. Size of each ~920 Mb (Windows 1.6Gb)

alexbaden:
Do we need that many caches? Is it advantageous to cache the L0 deps in their own conda cache vs the speedup we would get by using the general conda cache then install L0 deps? Also, what is the GitHub policy on storage for these caches?

Devjiu:
> Do we need that many caches? Is it advantageous to cache the L0 deps in their own conda cache vs the speedup we would get by using the general conda cache then install L0 deps? Also, what is the GitHub policy on storage for these caches?

Before this pr we used even more conda caches. (about 6 - number of caches = number of keys. I mean we created them but not used) 

Github cache is limited with 10Gb, So we simply need to fit in it. In common current caches can be found at https://github.com/intel-ai/hdk/actions/caches .
Currently we are caching m2 (~ 900 Mb not works for docker build, will fix later), docker image (~ 5.3Gb) and conda envs. 

Main profit of using so many caches is speed up all jobs.

.github/workflows/build.yml
```diff
@@ -80,6 +80,15 @@ jobs:
           sudo dpkg -i intel-igc-core_1.0.12812.24_amd64.deb intel-level-zero-gpu_1.3.25018.24_amd64.deb libigdgmm12_22.3.0_amd64.deb intel-igc-opencl_1.0.12812.24_amd64.deb
           conda install -n omnisci-dev -c conda-forge level-zero-devel pkg-config
 
+      - name: Save Conda env cache
+        if: steps.conda-cache.outputs.cache-hit != 'true'
```
leshikus, line 84: better to name ids for the same steps the same way imho; use either conda-cache or restore-conda-cache, not both in different places

Devjiu:
During current run - Approaching total cache storage limit (11.62 GB of 10 GB Used)
image

.github/workflows/pytest.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Set env context
         run: |
           echo CONDA_PATH=$CONDA >>$GITHUB_ENV
-          echo RUN_STAMP=${{ runner.os }}-${{ (env.cuda_compiler_version != 'None') && format('cuda{0}', env.cuda_compiler_version) || 'cpu' }} >>$GITHUB_ENV
+          echo RUN_STAMP=${{ runner.os }}-cpu >>$GITHUB_ENV
```
leshikus, line 19: both pytest.yml and modin.yml should be made a part of test.yml, thus the whole "Set env context" section will be reusable

leshikus (commented):
No critical problems, see two minor comments

.github/workflows/build-windows.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
       - name: Restore Conda env cache
         id: conda-cache
-        uses: actions/cache@v3
+        uses: actions/cache/restore@v3
```
leshikus, line 47: same here, better to have the same action name in all places; if you add "restore" here, add everywhere

.github/workflows/pytest.yml
Devjiu: will change it later in further prs

.github/workflows/build-windows.yml
Devjiu: done

.github/workflows/build.yml
Devjiu: done

kurapov-peter (approved):
Looks OK to me

PR #457, opened 2023-05-03
Fixed timestamp literals serialization.

AndreyPavlenko:
The timestamp literals must be converted to TimestampString, otherwise, Calcite fails with:

```java
java.lang.ClassCastException: class java.math.BigDecimal cannot be cast to class java.lang.Long (java.math.BigDecimal and java.lang.Long are in module java.base of loader 'bootstrap')
        at org.apache.calcite.rex.RexBuilder.clean(RexBuilder.java:1815) ~[calcite-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
        at org.apache.calcite.rex.RexBuilder.makeLiteral(RexBuilder.java:1593) ~[calcite-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
        at org.apache.calcite.rex.RexBuilder.makeLiteral(RexBuilder.java:1555) ~[calcite-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
        at org.apache.calcite.rel.externalize.MapDRelJson.toRex(MapDRelJson.java:497) ~[calcite-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
```

### DIFF
--- a/omniscidb/Calcite/java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java
+++ b/omniscidb/Calcite/java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java
@@ -16,15 +16,12 @@
 
 package org.apache.calcite.rel.externalize;
 
-import com.fasterxml.jackson.databind.type.TypeFactory;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.mapd.calcite.parser.MapDSqlOperatorTable;
 
-import org.apache.calcite.avatica.AvaticaUtils;
 import org.apache.calcite.avatica.util.TimeUnitRange;
 import org.apache.calcite.plan.RelOptCluster;
-import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.rel.RelCollation;
 import org.apache.calcite.rel.RelCollationImpl;
 import org.apache.calcite.rel.RelCollations;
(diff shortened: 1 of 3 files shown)


kurapov-peter (commented):
What is this for and why do we do it in java? Do we need the same transformation if python builders are used?

AndreyPavlenko:
> What is this for and why do we do it in java?

There is a bug in the Java part. Number literals are always converted to BigDecimal and the value is passed to makeLiteral() - https://github.com/intel-ai/hdk/blob/main/omniscidb/Calcite/java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java#L488. But, in case of timestamp, it's expected, that the value mut be either TimestampString, Calendar or Long. The last two are converted to TimestampString with millisecond precision - https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/rex/RexBuilder.java#L1827. That's why ClassCastException is happening here.

Another issue - the `precision` parameter is ignored in `toRex()` and handled incorrectly in `toJson()` and `parseLiteral()`

kurapov-peter (approved):
LGTM

omniscidb/Calcite/java/calcite/src/main/java/org/apache/calcite/rel/externalize/MapDRelJson.java
```diff
@@ -518,6 +533,73 @@ RexNode toRex(RelInput relInput, Object o) {
...
+    ts[9] = (char) ('0' + (day % 10));
+    ts[10] = ' ';
+    ts[11] = (char) ('0' + (hour / 10));
+    ts[12] = (char) ('0' + (hour % 10));
+    ts[13] = ':';
+    ts[14] = (char) ('0' + (minute / 10));
+    ts[15] = (char) ('0' + (minute % 10));
+    ts[16] = ':';
+    ts[17] = (char) ('0' + (second / 10));
+    ts[18] = (char) ('0' + (second % 10));
```
kurapov-peter, line 572: I'm a bit confused about all this part. Are there no methods in java to cut trailing zeroes or string builders?

AndreyPavlenko: There is the `StringBuilder` class, but I believe it would not be more efficient for this use case, because here know exactly the length of the string and positions of each character.

PR #459, opened 2023-05-04
Add ResultSetRegistry storage [7/7]

ienkovich:
The final part supports fixed-length arrays fetch in the ResultSetRegistry.

### DIFF
--- a/omniscidb/ResultSet/ResultSet.cpp
+++ b/omniscidb/ResultSet/ResultSet.cpp
@@ -890,14 +890,16 @@ bool ResultSet::isDirectColumnarConversionPossible() const {
 bool ResultSet::isZeroCopyColumnarConversionPossible(size_t column_idx) const {
   return query_mem_desc_.didOutputColumnar() &&
          query_mem_desc_.getQueryDescriptionType() == QueryDescriptionType::Projection &&
-         !colType(column_idx)->isVarLen() && appended_storage_.empty() && storage_ &&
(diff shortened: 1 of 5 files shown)


omniscidb/ResultSet/ResultSet.cpp
```diff
@@ -890,14 +890,16 @@ bool ResultSet::isDirectColumnarConversionPossible() const {
...
+         !colType(column_idx)->isVarLen() && !colType(column_idx)->isArray() &&
+         appended_storage_.empty() && storage_ &&
          (lazy_fetch_info_.empty() || !lazy_fetch_info_[column_idx].is_lazily_fetched);
 }
 
 bool ResultSet::isChunkedZeroCopyColumnarConversionPossible(size_t column_idx) const {
   return query_mem_desc_.didOutputColumnar() &&
          query_mem_desc_.getQueryDescriptionType() == QueryDescriptionType::Projection &&
-         !colType(column_idx)->isVarLen() && storage_ &&
+         !colType(column_idx)->isVarLen() && !colType(column_idx)->isArray() &&
```
kurapov-peter, line 901: Are arrays excluded because of the layout they have?

kurapov-peter (approved):

omniscidb/ResultSet/ResultSet.cpp
ienkovich: Yes. In ResultSet array value is actually a pointer to the data, so it cannot be zero-copy fetched.

PR #466, opened 2023-05-05
Add API to fetch a row from pyhdk.ExecutionResult.

### LINKED ISSUE #436: [PyHDK] Provide access to individual values of the execution result
We need a way to extract values from execution results by row and column indexes.

ienkovich:
This adds API to get rows from execution results. No support to examine imported tables for now.

Resolves #436

### DIFF
--- a/omniscidb/ResultSet/ResultSet.h
+++ b/omniscidb/ResultSet/ResultSet.h
@@ -251,10 +251,16 @@ class ResultSet {
 
   std::vector<TargetValue> getRowAt(const size_t index) const;
 
+  // TODO: among multiple rowAt versions only the following two accept actual row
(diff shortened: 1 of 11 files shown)


omniscidb/ResultSet/ResultSet.h
```diff
@@ -251,10 +251,16 @@ class ResultSet {
 
   std::vector<TargetValue> getRowAt(const size_t index) const;
 
+  // TODO: among multiple rowAt versions only the following two accept actual row
```
alexbaden, line 254: Yes, we should rename the old ones -- I like `getEntryAt`.

alexbaden (approved):

PR #471, opened 2023-05-09
Add a clang build

leshikus:
Also fix a conda forge CI


recipe/meta.yaml
```diff
@@ -1,6 +1,6 @@
 {% set version = "0.5.0" %}  # PEP 386
```
Garra1980, line 1: looks like non-related to clang build

leshikus: I intended to fix CI with this fix. To commit I need a greed CI

alexbaden (approved):

PR #473, opened 2023-05-10
Don't override CMAKE_INSTALL_PREFIX if it is already set for us

gshimansky:
This change is needed to implement conda build recipe on windows correctly (without ugly explicit list of every installed file).


gshimansky:
Does anyone know how to start CI here?

alexbaden:
I think this was a casualty of the GitHub outage from earlier. Usually closing/re-opening is enough to re-trigger once things stabilize.

kurapov-peter (approved):

PR #477, opened 2023-05-12
Temporary fallback for hash joins on L0

akroviakov:


kurapov-peter (approved):
LGTM, could you please squash it?

PR #479, opened 2023-05-12
Provide 'shared' implementations for count and sum functions

kurapov-peter:
Fixes these tests:
- Select.FilterAndSimpleAggregation
- Select.FilterAndGroupByMultipleAgg
- Select.GroupByPushDownFilterIntoExprRange
- Select.WatchdogTest

Those functions will have the `_shared` prefix once we enable the shared memory path.
A couple of notes:
1. I took a look into generating those functions by macros (specifically with m4), but it seems to me that typing them is less effort because of the subtle differences in the IR (mostly types casting).
2. Not all the platforms support atomics for, say, add i64, but I left it as is for now.
3. There's still at least one case that produces an incorrect result: `SELECT AVG(CAST(x AS FLOAT)) FROM test GROUP BY y;`

### DIFF
--- a/omniscidb/QueryEngine/Compiler/genx.ll
+++ b/omniscidb/QueryEngine/Compiler/genx.ll
@@ -60,6 +60,81 @@ define i64 @agg_count_skip_val(i64* %agg, i64 noundef %val, i64 noundef %skip_va
     ret i64 0
 }
 
+define i64 @agg_count_double_skip_val(i64* %agg, double noundef %val, double noundef %skip_val) {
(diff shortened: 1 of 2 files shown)


lmontigny (approved):

lmontigny:
Working on the SLM, I had to redefine in genx.ll the following methods as well:
`agg_min_double_skip_val`
`agg_max_float_skip_val`
`write_projection_int32`
write_projection_int64`

it has issue with the addrspace(4) casting as with `agg_sum_double_skip_val` previously.

PR #484, opened 2023-05-16
Disable most compiler warnings on windows

gshimansky:
This PR doesn't disable all of compiler warnings, but with warning level 0 compiler produces only the most serious ones. There are not many of these.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,14 @@ if (WIN32)
   # corresponding to the real C++ language standard version. Without this option Microsoft compiler
   # always defines __cplusplus to 199711L. A valid number is needed in llvm headers.
   add_compile_options(/Zc:__cplusplus)
+  # MSVC warnings management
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -38,6 +38,14 @@ if (WIN32)
   # corresponding to the real C++ language standard version. Without this option Microsoft compiler
   # always defines __cplusplus to 199711L. A valid number is needed in llvm headers.
   add_compile_options(/Zc:__cplusplus)
+  # MSVC warnings management
```
aregm, line 41: Can you please add a TODO comment that this needs to be fixed in the future?

alexbaden (approved):
Is the expectation that windows is always built from the root directory?

gshimansky:
Yes I moved warnings control from omniscidb level to the top because there are warning also happening in other places, e.g. python and HDK modules inherit compiler options from top level. Without moving this block I had contradicting options between top and omniscidb levels and MS compiler produced a separate warning that options are overridden.

PR #487, opened 2023-05-17
Add min/max implementations for L0 path

kurapov-peter:
The `cas` and `xchg` are there to be used by `get_value_group` implementation that will come later.
Fixes Select.GroupByPerfectHash and most of the Select.GroupByBaselineHash.


alexbaden (approved):
I have to say, I like the CUDA C++ definition of specialized device functions (like atomics) better. Is that a possibility for the `genx.ll` file in the future, either in DPC++ or some C/C++ API?

PR #495, opened 2023-05-22
Avoid overflow in get_matching_group_value.

ienkovich:
We use 32bit indices to access baseline hash tables, but its actual size might be over 4GB, so we shouldn't use 32bit offset values to access it. With this fix, we can successfully execute H2O GroupBy Q10 on a big dataset requiring a 2 billion entries baseline hash table.

### DIFF
--- a/omniscidb/QueryEngine/RuntimeFunctions.cpp
+++ b/omniscidb/QueryEngine/RuntimeFunctions.cpp
@@ -1187,7 +1187,7 @@ ALWAYS_INLINE GENERIC_ADDR_SPACE int64_t* get_matching_group_value(
     const T* key,
     const uint32_t key_count,
     const uint32_t row_size_quad) {
-  auto off = h * row_size_quad;
+  uint64_t off = static_cast<uint64_t>(h) * row_size_quad;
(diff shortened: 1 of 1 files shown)


kurapov-peter (commented):
Cuda has its own `get_matching_group_value` implementation that also uses `uint32_t` for offsets. Also, there are a bunch of similar functions such as `get_matching_group_value_columnar_slot` with the same logic. We probably need to modify them as well.

ienkovich:
> Cuda has its own `get_matching_group_value` implementation that also uses `uint32_t` for offsets. Also, there are a bunch of similar functions such as `get_matching_group_value_columnar_slot` with the same logic. We probably need to modify them as well.

I looked into `get_matching_group_value_columnar_slot` but didn't see any operation that might overflow. For CUDA, we don't allocate hash tables >2GB. I'm not sure it's relevant for CUDA, because the offset is not in bytes, it's in `sizeof(int64_t)`, so you need at least 32GB hash table to overflow. Do we have such big ones on CUDA?

kurapov-peter (approved):
> Do we have such big ones on CUDA?

No idea

PR #504, opened 2023-05-24
Remove reduntant error code store for no-interrupt case

kurapov-peter:


ienkovich:
Why is the error code store redundant? If dynamic interruption is disabled we still want to get error codes like overflow, etc.

kurapov-peter:
What happens is we already store it with `record_error_code` inside the row function, then at the upper level we load the value to check if an error occurred, and if yes we store it back again. I discovered this while triaging the GPU hanging issue on a specific query. I wasn't sure that's always the case (even though we seem to always do it unconditionally, see `createErrorCheckControlFlow`), so I created this draft to check with CI.

ienkovich:
I saw cases when `row_func` doesn't even get `error_code` but still can return a non-zero value. This happens when the number of matched rows exceeds `max_matched` value. In this case, we return `-pos` written later to `error_code`. I'm not sure if this is used in tests, but looks like a useful feature to interrupt and then continue execution from the last position.

PR #505, opened 2023-05-24
Use multifrag kernels on CPU for groupby with a big result.

ienkovich:
The basic idea is to not use multiple kernels when each kernel produces a hash table comparable in size to the whole input data. Then execution would go in a single thread and we would avoid a very costly reduction.

I'm sure my criteria are too strong and we might benefit from a single kernel in much more cases but this one at least would allow us to run queries like H2O Q10 on big data sets until we switch to better aggregation algorithms.


alexbaden (approved):
We discussed a bit in person yesterday -- this seems fine for now. I think it would be worthwhile to remove the extra group by buffers when we have baseline hash if we know the estimator query ran. The estimator query should tell us how many entries are in the buffer, making a group by buffer per fragment unnecessary. 
We discussed adding number of fragments to the heuristic but we aren't sure it would matter.

PR #511, opened 2023-06-05
Fix conda-forge build

leshikus:
Subj


alexbaden:
This is porting the changes made in hdk-feedstock back to our repo?

leshikus:
Yes, plus I've added some heuristics to guess a config name

alexbaden (approved):

PR #515, opened 2023-06-07
Add cpu-only testing for L0-enabled build

kurapov-peter:


omniscidb/Tests/GroupByTest.cpp
```diff
@@ -327,13 +313,13 @@ class LowCardinalityThresholdTest : public ::testing::Test {
...
-  for (auto dt : {ExecutorDeviceType::CPU, ExecutorDeviceType::GPU}) {
-    SKIP_NO_GPU();
-
+  config().exec.heterogeneous.allow_cpu_retry = true;
+  for (auto dt : testedDevices()) {
     auto result = run_multiple_agg(
         R"(select fl,ar,dep from low_cardinality group by fl,ar,dep;)", dt);
     EXPECT_EQ(result->rowCount(), config().exec.group_by.big_group_threshold);
   }
+  config().exec.heterogeneous.allow_cpu_retry = false;
```
ienkovich, line 322: Config is not going to be restored in case the test fails. We usually use `ScopeGuard` to restore config.

omniscidb/Tests/GroupByTest.cpp
```diff
@@ -389,12 +376,18 @@ int main(int argc, char** argv) {
   logger::LogOptions log_options(argv[0]);
   log_options.max_files_ = 0;  // stderr only by default
   desc.add(log_options.get_options());
+  desc.add_options()("cpu-only",
```
ienkovich, line 379: It would be better to use `ConfigBuilder` here and in all other places where it provides all the required options. That would enable the whole set of options, not only `cpu-only` one.

kurapov-peter:
> Why do we need CPU-only tests for L0 build?

This is to test the replacement of the default pyhdk package build with the L0-enabled one. I decided to go with a separate workflow to minimize testing disruption. We can replace the CPU build later if we find it more convenient.

omniscidb/scripts/conda/test.sh
leshikus, line 1: usage() function is no longer correct

leshikus (commented):
Having "cpu only" tests for l0 build seem excessive for me; they test the same things as "cpu only" tests for "cpu only" build. What am I missing here?

kurapov-peter:
> Having "cpu only" tests for l0 build seem excessive for me; they test the same things as "cpu only" tests for "cpu only" build. What am I missing here?

We have different binaries in L0-build and CPU cases. Again, I'm adding the new target so that the changes are sequential and we can be sure nothing is broken. Until we replace the CPU build packaged as pyhdk I'd keep the CPU workflow.

omniscidb/scripts/conda/test.sh
kurapov-peter: Thanks, fixed

ienkovich (approved):

Issue #521, opened 2023-06-12
Warning during compilation


Devjiu:
There are too many warnings generated during build. It makes cmake/make logs difficult to read and understand exact reason of an issue. 

One of possible solutions is to add job with `-Werror`. 

Also some warns are generated during cmake configuration, I am not sure it's covered with `Werror` option. 

Currently we have such errors:

during cmake:
```bash
CMake Warning (dev) at omniscidb/Tests/CMakeLists.txt:272:
  Syntax Warning in cmake code at column 204

  Argument not separated from preceding token by whitespace.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
This warning generated due to enabling of windows build and issued with cmd \s which have specific inner quotes handling. Maybe we can remove \s and have more difficult escaping for quotes from windows point of view, but it will not generate a warn? 

during make:
```bash
In file included from /localdisk/dmitriim/hdk/omniscidb/QueryEngine/RelAlgDagBuilder.h:36,
                 from /localdisk/dmitriim/hdk/omniscidb/QueryEngine/RelAlgExecutionUnit.h:31,
                 from /localdisk/dmitriim/hdk/omniscidb/QueryEngine/InputMetadata.h:22,
                 from /localdisk/dmitriim/hdk/omniscidb/QueryEngine/JoinFilterPushDown.h:22,
                 from /localdisk/dmitriim/hdk/omniscidb/QueryEngine/Descriptors/RelAlgExecutionDescriptor.h:19,
                 from /localdisk/dmitriim/hdk/omniscidb/QueryEngine/ArrowResultSet.h:20,
                 from /localdisk/dmitriim/hdk/omniscidb/Tests/TestHelpers.h:21,
                 from /localdisk/dmitriim/hdk/omniscidb/Tests/Quantile/QuantileCpuTest.cpp:19:
/localdisk/dmitriim/hdk/omniscidb/IR/Node.h: In member function 'virtual const std::string& hdk::ir::TranslatedJoin::getFieldName(size_t) const':
/localdisk/dmitriim/hdk/omniscidb/IR/Node.h:594:76: warning: no return statement in function returning non-void [-Wreturn-type]
  594 |   const std::string& getFieldName(size_t i) const override { CHECK(false); }
      |                                                                            ^
In file included from /localdisk/dmitriim/hdk/omniscidb/QueryEngine/RelAlgDagBuilder.h:36,
[...]

PR #523, opened 2023-06-12
[Warns] Reduce warnings amount

Devjiu:
This commit is an initial step to remove warnings from build. Currently
it's focused on linux build.

Always_inline attribute warning works incorrectly. In this case we want
to keep symbol and than inline it during target llvm ir optimisation.

See also: #521

### DIFF
--- a/omniscidb/QueryEngine/Optimization/AnnotateInternalFunctionsPass.h
+++ b/omniscidb/QueryEngine/Optimization/AnnotateInternalFunctionsPass.h
@@ -37,8 +37,6 @@ class AnnotateInternalFunctionsPass
                               llvm::CGSCCAnalysisManager& AM,
                               llvm::LazyCallGraph& CG,
                               llvm::CGSCCUpdateResult& UR) {
-    bool updated_function_defs = false;
-
 #if 0
     for (llvm::LazyCallGraph::Node& N : InitialC) {
       llvm::Function* Fn = &N.getFunction();
@@ -57,7 +55,6 @@ class AnnotateInternalFunctionsPass
       }
       if (isInternalStatelessFunction(fcn->getName()) ||
           isInternalMathFunction(fcn->getName())) {
-        updated_function_defs = true;
         std::vector<llvm::Attribute::AttrKind> attrs{llvm::Attribute::NoFree,
                                                      llvm::Attribute::NoSync,
                                                      llvm::Attribute::NoUnwind,
@@ -68,7 +65,6 @@ class AnnotateInternalFunctionsPass
           fcn->addFnAttr(attr);
         }
       } else if (isReadOnlyFunction(fcn->getName())) {
-        updated_function_defs = true;
         fcn->addFnAttr(llvm::Attribute::ReadOnly);
       }
     }
--- a/omniscidb/Tests/ArrowStorageTest.cpp
+++ b/omniscidb/Tests/ArrowStorageTest.cpp
@@ -13,6 +13,7 @@
  */
 
 #include "ArrowStorage/ArrowStorage.h"
+#include "Shared/ArrowUtil.h"
 
 #include "TestHelpers.h"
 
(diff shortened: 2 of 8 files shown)


Devjiu:
Replacing with FORCE_INLINE does not help for:
```bash
/localdisk/dmitriim/hdk/omniscidb/Utils/ExtractFromTime.cpp:199:30: warning: 'always_inline' function might not be inlinable [-Wattributes]
  199 | ALWAYS_INLINE DEVICE int64_t extract_week(const int64_t timeval) {
```

In this case looks like we should use simple inline. Is this an acceptable approach?

omniscidb/Tests/ArrowStorageTest.cpp
```diff
@@ -736,11 +736,11 @@ TEST_F(ArrowStorageTest, ImportArrowTable) {
   ArrowStorage storage(TEST_SCHEMA_ID, "test", TEST_DB_ID, config_);
   auto col_a = std::make_shared<arrow::Field>("A", arrow::null());
   auto schema = arrow::schema({col_a});
-
   std::shared_ptr<arrow::Array> empty_array;
   arrow::NumericBuilder<arrow::FloatType> float64_builder;
-  float64_builder.AppendValues({});
-  float64_builder.Finish(&empty_array);
+  // auto status = float64_builder.AppendValues({});
+  ASSERT_TRUE(float64_builder.AppendValues({}).ok());
```
alexbaden, line 742: There is a macro you can use here -- `ARROW_THROW_NOT_OK`  I believe.

omniscidb/QueryEngine/Optimization/AnnotateInternalFunctionsPass.h
```diff
@@ -37,7 +37,7 @@ class AnnotateInternalFunctionsPass
                               llvm::CGSCCAnalysisManager& AM,
                               llvm::LazyCallGraph& CG,
                               llvm::CGSCCUpdateResult& UR) {
-    bool updated_function_defs = false;
+    [[maybe_unused]] bool updated_function_defs = false;
```
alexbaden, line 40: It looks like this is actually not used? @lmontigny

omniscidb/IR/Node.h
```diff
@@ -591,7 +591,11 @@ class TranslatedJoin : public Node {
     CHECK(false);
     return nullptr;
   }
-  const std::string& getFieldName(size_t i) const override { CHECK(false); }
+  const std::string& getFieldName(size_t i) const override {
+    CHECK(false);
+    // just a stub, shouldn't be reached.
+    return op_type_;
```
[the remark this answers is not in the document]
ienkovich: Returning a reference to a local object would produce another warning though. I wonder if we can modify `CHECK` in such a way that the compiler would understand what we actually call `abort`. Or maybe use some new macro like `FAIL` to avoid unreachable return statements.

omniscidb/QueryEngine/Optimization/AnnotateInternalFunctionsPass.h
lmontigny: I confirm, it's not used

Devjiu:
> Can you also remove the FORCE and ALWAYS inline changes for now? I think that will require more thought. Let's tackle the other warnings now and look at inline warnings later.

Sure, It was an initial approach. AFAIU Force inline removes symbol for this function and we can't call it on runtime, so many tests are failed with forced. Looks like warning generated by gcc is incorrect for our case - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55830 . So I just suppressed it.

CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ project(
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_SUPPRESS_DEVELOPER_WARNINGS 1 CACHE INTERNAL "No dev warnings")
```
alexbaden, line 18: I was able to fix the cmake warning here: https://github.com/intel-ai/hdk/pull/531

omniscidb/IR/Node.h
```diff
@@ -591,7 +591,14 @@ class TranslatedJoin : public Node {
     CHECK(false);
     return nullptr;
   }
-  const std::string& getFieldName(size_t i) const override { CHECK(false); }
+  const std::string& getFieldName(size_t i) const override {
+    CHECK(false);
+    // Check(false) macro have abort in Logger d-tor,
```
alexbaden, line 596: Let's just return an empty string, nobody will ever actually access it (b/c of the CHECK), and that is consistent with the rest of the code base.

alexbaden (changes requested):

CMakeLists.txt
Devjiu: Cool) I also tried in the same way, but didn't assume that the extra space would resolve it.

omniscidb/IR/Node.h
ienkovich: You can also throw an exception instead of return to resolve the warning.

alexbaden: Ok, I guess we can leave the abort if it builds fine w/ gcc and clang.

alexbaden (changes requested):

alexbaden:
@Devjiu can you address https://github.com/intel-ai/hdk/pull/523#discussion_r1237796558 so we can merge this? It would be nice to suppress some of these warnings.

alexbaden (approved):

kurapov-peter (approved):

PR #524, opened 2023-06-13
Allow custom Calcite (java) / HDK log paths

### LINKED ISSUE #522: Running tests puts `${sys:MAPD_LOG_DIR}` directories in test folder
It appears that the environment variable `MAPD_LOG_DIR` set here https://github.com/intel-ai/omniscidb/blob/jit-engine/Calcite/CMakeLists.txt#L30 is not being picked up by the log4j properties file(s) https://github.com/intel-ai/omniscidb/blob/jit-engine/Calcite/java/calcite/src/main/resources/log4j2.properties. 

To reproduce, build as normal, then enter `build/Tests` and run `ArrowBasedExecuteTest --gtest_filter=Select.GroupBy` (keeps the tests brief).

@vlad-penkin Ilya suggested you might have some ideas about how to debug?

alexbaden:
Allows a user to customize the location for all log directories at runtime. 

For now I have put the log path in Config Options under Debug. We should move the log options to Config eventually. I probably could have named it "java log path" but I don't see a need to keep separate paths in the future, and would like to limit the amount of renaming needed. 

I also added a unit test to pyhdk.

Closes #522

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -63,4 +63,4 @@ jobs:
 
       - name: Run pytest
         run: |
-          $CONDA/bin/conda run -n ${{ env.CONDA_ENV }} pytest python/tests/ --ignore=python/tests/modin
+          $CONDA/bin/conda run -n ${{ env.CONDA_ENV }} pytest -s python/tests/ --ignore=python/tests/modin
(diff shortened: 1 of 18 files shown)


ienkovich (approved):
I wonder if it would be better to run each Python suite (or some of them) separately. With the current approach, we initialize come of HDK components only once per all tests and therefore cannot test different configs. And we encounter some bugs reproducible only with non-default config (e.g. the one used by Modin).

PR #534, opened 2023-06-19
Enable shared memory by default

kurapov-peter:
The PR adds proper code generation for Intel GPUs.

Notes:
* The shared buffer is allocated statically within the `genx.ll` module and its size is not checked for now.
* L0 backend now uses smem by default.
* There was an issue with cgen: `emitCall` would use the wrong runtime module. This is now fixed by patching the module triple upon extension module load.
* We currently produce some redundant address space casts between generic and shared-local. Those produce actual runtime checks and affect perf. I left the warnings in the log.

### DIFF
--- a/omniscidb/QueryEngine/Compiler/genx.ll
+++ b/omniscidb/QueryEngine/Compiler/genx.ll
@@ -1,3 +1,4 @@
+target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f32:32:32-f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64"
 target triple = "spir64-unknown-unknown"
 
 declare i64 @__spirv_BuiltInGlobalInvocationId(i32 %dimention)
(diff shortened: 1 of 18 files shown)


Garra1980:
Any specific tests for sh mem or existing set is enough?


> * We currently produce some redundant address space casts between generic and shared-local. Those produce actual runtime checks and affect perf. I left the warnings in the log.

Is it worth to be handled as a separate issue?

kurapov-peter:
The pass rate is now decent, so this is ready for review.

omniscidb/QueryEngine/Compiler/genx.ll
```diff
@@ -44,41 +60,52 @@ define i8 @thread_warp_idx(i8 noundef %warp_sz) {
...
+    %ld = load i32, i32 addrspace(4)* %agg
+    %old = atomicrmw add i32 addrspace(4)* %agg, i32 1 monotonic
+    ret i32 %old
+}
+
+define i64 @agg_count_skip_val_shared(i64 addrspace(4)* %agg, i64 noundef %val, i64 noundef %skip_val) {
     %no_skip = icmp ne i64 %val, %skip_val
     br i1 %no_skip, label %.noskip, label %.skip
 .noskip:
-    %old = call i64 @agg_count(i64* %agg, i64 %val)
```
alexbaden, line 57: What happens if `shared_mem` is turned off? What function implementation do we fall back to?

alexbaden (approved):

omniscidb/QueryEngine/Compiler/genx.ll
kurapov-peter: Just noticed the comment. I haven't tested it, but what should happen is it should emit the `agg_count` from the "cpu" runtime module instead. BTW, I'm not sure if we need the option to disable shared memory at all :)

PR #538, opened 2023-06-21
Support partitioned aggregation.

ienkovich:
This patch adds a simple version of partitioned aggregation. Only single-step hash partitioning is supported with quite simple heuristics to cover big outputs with many small groups. This can be further improved but need more cases to tune heuristics. 

The current implementation compares the estimated output buffer size with the input size trying to detect big buffers whose reduction wouldn't allow scaling. If the condition is fulfilled, then a replay with partitioning is requrest.

Partitioned aggregation goes in multiple steps.

Step 0 (optional). Materialize data to partition. This is to avoid costly operations executed multiple times on shuffle stages. Again, there is no good heuristics on what operation is simple enough to be included in the shuffle, so for now, I simply execute everything except simple projections.

Step 1. Building histogram for partitions. The goal here is to compute the sizes of future partitions and also to determine offsets to be used by each input fragment in output partitions (so that all kernels could write into the same buffers with no additional sync). This is done using a new node `Shuffle`. The node has group expressions to be used for partitioning and target expressions to be written to output partitions. For this step, it is just `COUNT(*)` expression. On output, we want to get simple arrays with partition sizes, so I force a keyless perfect hash table for memory layout providing exactly that case. There is a special reduction added in `Executor` for such kernel to compute partial sums of obtained partition sizes to be later used as offsets in output buffers. As a result, for `N` input partitions we have `N + 1` output partitions. The output partition with the number K would hold output offsets for the input partition with the same number. I also use some nasty trick to zero-copy transform perfect hashes into projections so that we could zero-copy fetch them later. Probably, it's not performance critical and we could simply create new projection result sets to hold the final data (perfect hash tables are not working because we don't want to skip empty rows)

Step 2. Partitioning. Here we use the same node but use column references in target expressions for all columns we need in the output. Also, we add the result of Step 1 as the second input table. The output is a projection. The biggest trick here is to pre-allocate result sets of the required size for all partitions and generate code that could write to multiple result sets at once. Currently, I pre-allocate all required buffers in Executor (and I'm thinking about moving it to QueryMemoryInitializer), then store for each output partition I store all column pointers in arrays (vectors), then build a new vector with pointers to all built arrays. As a result, I get `int8_t ***` pointer, and using a partition number and then a target number allows me to get the required output column pointer. This pointer is then passed to the kernel instead of the hash table pointer. All kernels get the same pointer but use a proper fragment (with the same number as the processed one) to use their own offsets for output columns. Some changes to code generation were introduced to allow each target to use its own output buffer instead of applying a known offset to a single base value.

Step 3. Aggregation. This is a pretty normal aggregation with slight modifications. First, it has an output entry count hint to avoid new estimation requests and use a sane output buffer size. Second, it allows us to completely avoid a reduction, unless the next operation or config says otherwise. New option to allow multi-fragment execution results (final result, not intermediate) is not disabled by default because such results don't support import to Arrow yet. Also, enabling it by default would most probably break a lot of our tests. I intend to support all public PyHDK API for multi-fragment result set tables and then enable this option by default (leaving it off for testing).

With default parameters, this partitioning would work only for H2O GroupBy Q10 on medium and big data sets. The gain depends on data size and fragment size. For big data size and the default fragment size, the preliminary boost is x5-6. Less for medium data size. Didn't measure on a clean machine yet.

There is still room for improvements. E.g. avoid metadata computation for partitioned data and use different sizes for final aggregation to avoid big partitions affecting all buffer sizes. Popular radix partitioning also assumes using multiple partitioning steps to avoid TLB pressure and get smaller partitions (to have resulting hash table fitting L1/L2 caches).

### DIFF
--- a/omniscidb/QueryEngine/Execute.cpp
+++ b/omniscidb/QueryEngine/Execute.cpp
@@ -70,6 +70,7 @@
 #include "QueryEngine/SpeculativeTopN.h"
 #include "QueryEngine/StringDictionaryGenerations.h"
 #include "QueryEngine/Visitors/TransientStringLiteralsVisitor.h"
+#include "ResultSet/ColRangeInfo.h"
 #include "Shared/checked_alloc.h"
 #include "Shared/funcannotations.h"
 #include "Shared/measure.h"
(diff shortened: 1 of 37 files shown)


omniscidb/QueryEngine/Execute.cpp
```diff
@@ -1360,6 +1361,76 @@ ResultSetPtr Executor::reduceSpeculativeTopN(
...
+                                    data_mgr_,
+                                    blockSize(),
+                                    gridSize());
+    proj_rs->allocateStorage(buf, {});
+    results.push_back(proj_rs);
+    buffers.push_back(reinterpret_cast<int64_t*>(buf));
+  }
+
+  bzero(buffers[0], sizeof(int64_t) * parts);
+  for (size_t i = 2; i < buffers.size(); ++i) {
```
alexbaden, line 1425: Why does this index start at 2?

alexbaden (approved):
Looks nice. I particularly like the way things are implemented in RelAlgExecutor. We have not really done stepwise execution inside a single RelAlg node before, but it seems to work very well here. 

On the executor side it looks good too, but I looking at the creation of the partition buffers and shuffle reductions, we are able to re-use existing data structures and codegen but with a lot of hacks (for example -- the data mgr unpin routine in the unique ptr, the custom query memory descriptor, and custom allocation routine in execute work unit -- for this last one, I wonder if we're getting other output buffers too. Perhaps we need them?). Regardless, this is fine for now but it does make me wonder about refactors we could do that would make it easier to support using our baseline data structures, execution paths, and codegen for these types of operations. What do you think?

omniscidb/QueryEngine/Execute.cpp
ienkovich: The first one is all zeroes, so no need to add them to the second one, and therefore start from the third one.

ienkovich:
For shuffle reduction, when we compute partial sums, we can re-use existing infrastructure to merge two result sets. It's simply a perfect hash with a count aggregate. The big difference is that we need to get N + 1 result sets in the end, so it's just the way how the reduction is applied to result sets, We can run the reduction sequence RS1->RS2->...->RSN to get the required partial sums. I just went the easy way and didn't want time spent on module generation or interpretation when there is a simple and efficient loop to compute everything.

As for shuffle buffers creation and passing them to the module, I think this is the weakest part of the patch. I don't like the forced query memory descriptors and result set creation in the executor. But I don't see a better way to do it yet. I was thinking about moving this part to QueryMemoryInitializer. The problem with going the regular way is that the work unit is both a projection (by QueryDescriptionType) and a group by (because we have group by exprs in the unit), so multiple unobvious fixes are required along the way. I tried to introduce a new QueryDescriptionType::Shuffle, but it didn't actually give desired simplicity and also required supporting the new type all over the codegen. I still think it should go through QueryMemoryInitializer but more time is required to find the best way to do it. Maybe an attempt to extend this to joins would give a better understanding of how it should look at the end.

I think we still create unused output buffers in QueryExecutionContext. These should be with 1 entry count only (due to the query memory descriptor built for a shuffle in MemoryLayoutBuilder), but it is still a clumsiness to be fixed. I guess moving shuffle buffer initialization to QueryMemoryInitializer would fix this problem as well, But it's not clear how to share buffers between all contexts then. Would using SharedKernelContext in QueryMemoryInitializer be reasonable?

I created two new issues to address these problems.

PR #553, opened 2023-06-27
Add non-lazy data import.

ienkovich:
Add a new option that allows to avoid data fetch costs on execution moving it to data import. This option is for benchmarking only when we want to split data import/transformation time and execution time. This can lead to an overall workload execution time increase in the case of unused columns in imported data.

### DIFF
--- a/omniscidb/ArrowStorage/ArrowStorage.cpp
+++ b/omniscidb/ArrowStorage/ArrowStorage.cpp
@@ -715,7 +715,8 @@ void ArrowStorage::appendArrowTable(std::shared_ptr<arrow::Table> at, int table_
         switch (col_arr->type()->id()) {
           case arrow::Type::STRING:
             // if the dictionary has already been materialized, append indices
-            if (!config_->storage.enable_lazy_dict_materialization ||
+            if (config_->storage.enable_non_lazy_data_import ||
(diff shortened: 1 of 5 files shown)


omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -715,7 +715,8 @@ void ArrowStorage::appendArrowTable(std::shared_ptr<arrow::Table> at, int table_
         switch (col_arr->type()->id()) {
           case arrow::Type::STRING:
             // if the dictionary has already been materialized, append indices
-            if (!config_->storage.enable_lazy_dict_materialization ||
+            if (config_->storage.enable_non_lazy_data_import ||
```
alexbaden, line 718: Is this necessary simply to force non-lazy data import in all cases when non lazy data import is enabled in config (vs just lazy dict materialization), or is there a codepath here that is important if we want non-lazy data import and lazy dict materialization is already disabled?

ienkovich: I did this change assuming that lazy dictionaries are going to be enabled by default soon and I would like to disable all lazy import features with a single flag in benchmarks.

alexbaden (approved):

PR #556, opened 2023-06-28
Add missing destructor

kurapov-peter:
Well... happens, I guess.

### DIFF
--- a/omniscidb/L0Mgr/L0Mgr.cpp
+++ b/omniscidb/L0Mgr/L0Mgr.cpp
@@ -114,6 +114,10 @@ ze_command_list_handle_t L0CommandList::handle() const {
 
 L0CommandList::~L0CommandList() {
   // TODO: maybe return to pool
+  auto status = zeCommandListDestroy(handle_);
+  if (status) {
+    std::cerr << "Non-zero status for command list destructor" << std::endl;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


omniscidb/L0Mgr/L0Mgr.cpp
```diff
@@ -114,6 +114,10 @@ ze_command_list_handle_t L0CommandList::handle() const {
 
 L0CommandList::~L0CommandList() {
   // TODO: maybe return to pool
+  auto status = zeCommandListDestroy(handle_);
+  if (status) {
+    std::cerr << "Non-zero status for command list destructor" << std::endl;
```
ienkovich, line 119: Why not LOG(ERROR)?

kurapov-peter: No particular reason, just to stay consistent.

ienkovich: I see. I still think moving from `std::err` to `LOG(ERROR)` in this file would be the right thing to do. In tests, ERROR log should be printed to stderr anyway.

ienkovich (approved):

omniscidb/L0Mgr/L0Mgr.cpp
kurapov-peter: Done

PR #560, opened 2023-06-29
Fix docker uuid

leshikus:
Security fix - do not allow a docker user access ghrunner resources

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -71,7 +71,9 @@ jobs:
       - name: Start the container
         run:  |
           mkdir -p build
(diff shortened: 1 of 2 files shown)


Devjiu (commented):
I still don't understand what security issue is fixed, so I can't approve this.

Devjiu (approved):
Discussed the issue offline. This does not eliminate all possible problems, but it prevents the simplest exploits.
There is another approach, but it requires more work with almost the same result.
So we decided to keep this one to solve the security problem.

leshikus:
@alexbaden could you please take a look as well; @Garra1980 asked CI patches to be reviewed by you or Ilya

alexbaden (approved):

PR #564, opened 2023-06-30
Add support for bitwise operations.

ienkovich:
Covers C++ and PyHDK API but not Calcite.

Related to #542.


omniscidb/QueryBuilder/QueryBuilder.h
```diff
@@ -156,6 +156,20 @@ class BuilderExpr {
   BuilderExpr logicalAnd(const BuilderExpr& rhs) const;
   BuilderExpr logicalOr(const BuilderExpr& rhs) const;
 
+  BuilderExpr bwAnd(const BuilderExpr& rhs) const;
+  BuilderExpr bwAnd(int val) const;
+  BuilderExpr bwAnd(int64_t val) const;
+
+  BuilderExpr bwOr(const BuilderExpr& rhs) const;
+  BuilderExpr bwOr(int val) const;
+  BuilderExpr bwOr(int64_t val) const;
```
kurapov-peter, line 165: Should we use explicit bit width everywhere? I'd expect either `int32_t` with `int64_t` or `int`  with `long long`; the latter should be incorrect in our case.

python/pyhdk/hdk.py
```diff
@@ -993,6 +993,121 @@ def logical_or(self, value):
         """
         pass
 
+    def bw_and(seld, value):
```
kurapov-peter, line 996:
```suggestion
    def bw_and(self, value):
```

python/pyhdk/hdk.py
```diff
@@ -993,6 +993,121 @@ def logical_or(self, value):
...
+          expr_1: INT64
+        Data:
+        0
+        2
+        2
+        0
+        """
+        pass
+
+    def bw_or(seld, value):
```
kurapov-peter, line 1026:
```suggestion
    def bw_or(self, value):
```

python/pyhdk/hdk.py
```diff
@@ -993,6 +993,121 @@ def logical_or(self, value):
...
+          expr_1: INT64
+        Data:
+        5
+        3
+        3
+        5
+        """
+        pass
+
+    def bw_xor(seld, value):
```
kurapov-peter, line 1056:
```suggestion
    def bw_xor(self, value):
```

python/pyhdk/hdk.py
```diff
@@ -993,6 +993,121 @@ def logical_or(self, value):
...
+          expr_1: INT64
+        Data:
+        5
+        1
+        1
+        5
+        """
+        pass
+
+    def bw_not(seld):
```
kurapov-peter, line 1086:
```suggestion
    def bw_not(self):
```

kurapov-peter (changes requested):
Looks good, some typos

omniscidb/QueryEngine/CalciteDeserializerUtils.h
```diff
@@ -74,6 +74,18 @@ inline hdk::ir::OpType to_sql_op(const std::string& op_str) {
...
+  }
+  if (op_str == std::string("BW_OR")) {
+    return hdk::ir::OpType::kBwOr;
+  }
+  if (op_str == std::string("BW_XOR")) {
+    return hdk::ir::OpType::kBwXor;
+  }
+  if (op_str == std::string("BW_NOT")) {
+    return hdk::ir::OpType::kBwNot;
+  }
```
AndreyPavlenko, line 88: Calcite has support for BIT_AND/OR/XOR. Support for BIT_NOT is added here - https://github.com/intel-ai/hdk/pull/580. To be compatible with Calcite, the operation name prefix should be changed to `BIT_`.

ienkovich: Added both options

omniscidb/QueryBuilder/QueryBuilder.h
ienkovich: `int` and `int64_t` overloads match all other similar functions. We can use `int32_t` and `int64_t` instead, but it should be done globally then, not only for bitwise operations.

Issue #567, opened 2023-07-04
[Perf][Bench] Poor performance of join


Devjiu:
In several reference requests join queries have similar issues. 
In case of big tables main performance drop is issued by fetchChunks stage.
In current case AFAIU the slowest stage is compileWorkUnit, 
```js 
198 ms total duration for executeRelAlgQuery
        197 ms start(0ms) executeWorkUnit RelAlgExecutor.cpp:1389
          119 ms start(0ms) compileWorkUnit NativeCodegen.cpp:1403
          0 ms start(120ms) fetchChunks Execute.cpp:3090
          8 ms start(121ms) collectAllDeviceResults Execute.cpp:2550
          0 ms start(195ms) fetchChunks Execute.cpp:3090
          1 ms start(196ms) executePlan Execute.cpp:3462
```
60% of total execution time was spent in compileWorkUnit. 

Pandas in this case executes query for ~5 ms vs ~240 ms HDK for single query. 

Looks like the most significant part of compileWorkUnit is
```js
  New thread(1)
    15 ms start(0ms) initHashTableForDevice BaselineJoinHashTable.cpp:562
      0 ms start(0ms) initHashTableOnCpuFromCache BaselineJoinHashTable.cpp:980
      0 ms start(0ms) initHashTableOnCpuFromCache BaselineJoinHashTable.cpp:980
      15 ms start(0ms) initHashTableOnCpu BaselineHashTableBuilder.h:193
        1 ms start(0ms) CPU Baseline-Hash: init_baseline_hash_join_buff_32 BaselineHashTableBuilder.h:227
  End thread(1)
  New thread(386)
    62 ms start(0ms) initHashTableForDevice BaselineJoinHashTable.cpp:562
      0 ms start(0ms) initHashTableOnCpuFromCache BaselineJoinHashTable.cpp:980
      0 ms start(0ms) initHashTableOnCpuFromCache BaselineJoinHashTable.cpp:980
      62 ms start(0ms) initHashTableOnCpu BaselineHashTableBuilder.h:193
        0 ms start(0ms) CPU Baseline-Hash: init_baseline_hash_join_buff_32 BaselineHashTableBuilder.h:227
        0 ms start(11ms) CPU Baseline-Hash: Additional Buffers init_hash_join_buff BaselineHashTableBuilder.h:309
  End thread(386)
```
[...]

PR #568, opened 2023-07-04
[Bench] Reference for perfomance analysis

Devjiu:
This is an example that compares HDK with Pandas performance. Can be modified by increasing N (originally was 10).

See also: #567


Devjiu:
no longer needed

PR #591, opened 2023-07-20
Add cardinality API to QueryBuilder.

ienkovich:


omniscidb/QueryBuilder/QueryBuilder.cpp
```diff
@@ -1587,6 +1587,30 @@ BuilderExpr BuilderExpr::at(int64_t idx) const {
...
+  if (!expr_->type()->isArray()) {
+    throw InvalidQueryError() << "Non-array type used for CARDINALITY operator: "
+                              << expr_->type()->toString();
+  }
+
+  if (expr_->is<Constant>()) {
+    if (expr_->as<Constant>()->isNull()) {
+      return builder_->nullCst(builder_->ctx_.int32());
+    }
+    return builder_->cst(static_cast<int32_t>(expr_->as<Constant>()->valueList().size()),
```
kurapov-peter, line 1600: Why is the size casted to signed int here but not below (line 1606)?

kurapov-peter (approved):

omniscidb/QueryBuilder/QueryBuilder.cpp
ienkovich: This is because `hdk::ir::Type::size` returns `int`, not `size_t`.

PR #595, opened 2023-07-24
Support UNNEST in projections.

ienkovich:
This PR adds UNNEST support for projection queries (currently, we support it in GROUP BY only). Implementation is quite simple, based on target_exprs holding UNNEST expressions. This is not generic enough and basically makes UNNEST projections to be executed as a separate step, preventing it from merging with other nodes. I think we can get a better solution by introducing a new set of expressions specifically for unnesting in the execution unit, but this work might become a much bigger project than I'm willing to have for my goals.

For now, I need this simple solution to combine it with my ongoing work on TopK aggregate, which would produce an array and require the following UNNEST projection to cover H2O GroupBy Q8. This will let us avoid Window Functions for this query and make its execution scalable (now, we run it in a single thread and have an additional projection to get a single fragment prior to window function execution).

### DIFF
--- a/omniscidb/IR/Node.cpp
+++ b/omniscidb/IR/Node.cpp
@@ -6,7 +6,9 @@
  */
 
 #include "Node.h"
+#include "ExprCollector.h"
 #include "InputRewriter.h"
+#include "UnnestDetector.h"
 
 namespace hdk::ir {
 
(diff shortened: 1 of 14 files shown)


alexbaden (approved):
Looks good, though you may have a problem w/ L0 tests.

I am curious what happens on GPU with the output slots - IIRC we use an atomic to get the next output slot. So, theoretically UNNEST would give you a different output ordering every time (whereas CPU should be more or less consistent, particularly if you only have one fragment).

ienkovich:
> I am curious what happens on GPU with the output slots - IIRC we use an atomic to get the next output slot.

I increment this atomic counter by an array size and then use an additional non-atomic counter to output all rows into the reserved space. That should give me an adequate order.

Using new features on CPU only first would also be acceptable for me. With the following features port for other targets.

PR #597, opened 2023-07-25
[Werror] Add jobs that checks warnings

Devjiu:
This commit adds jobs and cmake options to verify clang/gcc builds with Werror flag.


alexbaden:
My position on this is unchanged - it is cheaper and easier to fix warnings as we notice them vs having more CI jobs, which add complexity and do have a cost in time if not in money.

Devjiu:
Decided to leave it as is. Warnings will be fixed as we notice them.

PR #599, opened 2023-07-27
Fix default compilation options construction in arrow sql runner implemenation

kurapov-peter:
The runner would create an incorrect traits descriptor (i.e. with cuda-specific settings).


kurapov-peter:
> Do you need a similar change for `pyhdk._sql.RelAlgExecutor.execute`?

Yep, good catch. I think I should redo it a bit. Maybe make the platform a required parameter for `defaults()`. Otherwise, it gets messy.

PR #601, opened 2023-07-31
[Namings] Update job names

Devjiu:
This commit unified gcc/clanf build/tests.


.github/workflows/test-docker.yml
```diff
@@ -65,7 +80,7 @@ jobs:
           docker exec -u ghrunner hdk-build.${{ inputs.name }} tar -zxf /_work/build.tgz -C /_work/
 
       - name: Sanity test
-        if: inputs.name == 'cuda'
+        if: inputs.name == 'cuda'|| inputs.name == 'clang-cpu'
```
kurapov-peter, line 83: Seems like this didn't run for the clang-built HDK, huh?

Devjiu: yep

alexbaden:
@leshikus can you comment?

.github/workflows/test-docker.yml
```diff
@@ -37,10 +37,25 @@ jobs:
...
+      - name: Choose Dockerfiles
+        if: inputs.name == 'cuda'
+        run: |
+          echo Additional_Dockerfile=docker/Dockerfile.cuda >>$GITHUB_ENV
+      
+      - name: Choose Dockerfiles
+        if: inputs.name == 'l0'
+        run: |
+          echo Additional_Dockerfile=docker/Dockerfile.l0 >>$GITHUB_ENV
+
```
leshikus, line 54: I believe this overcomplicates things; we always can name the dockerfile as we like, so there is no need to have this additional "switch" to choose the name

leshikus (approved):
LGTM

PR #610, opened 2023-08-02
Generalize compile options creation

kurapov-peter:
This is an attempt to prevent issues like those in https://github.com/intel-ai/hdk/pull/599#pullrequestreview-1550317117. The PR adds CompilationOptionsBuilder and uses it in the python execute API.

This is somewhat similar to what we've done to execution options earlier. The goal would be to move to the new builder everywhere and remove all the direct options modification*, although the config might not be always present. We could use another implementation for such cases. The builder would also be responsible for any validation needed. For example, to check that `device_type` and the `codegen_traits_desc` are compliant.

*For example, things like ` c_co.get().allow_lazy_fetch = kwargs.get("enable_lazy_fetch", config.rs.enable_lazy_fetch)` should modify the config itself and everything else uses the config as a single source of truth.

Thoughts?

### DIFF
--- a/omniscidb/CudaMgr/CudaMgr.h
+++ b/omniscidb/CudaMgr/CudaMgr.h
@@ -164,6 +164,8 @@ class CudaMgr : public GpuMgr {
     return getMinNumMPsForAllDevices();
   };
 
+  bool hasFP64Support() const override { return isArchPascalOrLater(); }
+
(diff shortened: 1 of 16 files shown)


kurapov-peter:
Ugh. I was trying to reproduce the problem locally, but `QueryBuilderTest` fails with an assertion for me. Updated conda, clean build - still failing.

PR #613, opened 2023-08-02
Remove exceptions from decimal validator message

alexbaden:
Using exceptions results in substantial performance loss in tight loops, where this code is often used.

### DIFF
--- a/omniscidb/DataMgr/FixedLengthEncoder.h
+++ b/omniscidb/DataMgr/FixedLengthEncoder.h
@@ -91,7 +91,6 @@ class FixedLengthEncoder : public Encoder {
           for (size_t i = range.begin(); i < range.end(); i++) {
             if (data[i] != inline_null_value<V>()) {
               if (!fixlen_array || data[i] != inline_null_array_value<V>()) {
-                decimal_overflow_validator_.validate(data[i]);
                 min = std::min(min, data[i]);
                 max = std::max(max, data[i]);
               }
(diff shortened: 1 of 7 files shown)


omniscidb/DataMgr/FixedLengthEncoder.h
```diff
@@ -91,9 +91,10 @@ class FixedLengthEncoder : public Encoder {
           for (size_t i = range.begin(); i < range.end(); i++) {
             if (data[i] != inline_null_value<V>()) {
               if (!fixlen_array || data[i] != inline_null_array_value<V>()) {
-                decimal_overflow_validator_.validate(data[i]);
-                min = std::min(min, data[i]);
-                max = std::max(max, data[i]);
+                if (decimal_overflow_validator_.validate(data[i])) {
```
ienkovich, line 94: Do we just ignore out-of-range decimal values now or do we assume it's checked on the arrow decimal conversion? Shouldn't this validation always succeed?

alexbaden: Well, I am skeptical we ever hit this branch but can't guarantee it. So, we ignore out of range values. I suppose this could pose a problem later, and maybe we should just remove this code altogether and make sure we validate the decimal range on import. Alternatively we could try aborting with an error message (or perhaps an exception) and not specify whether the decimal is above or below the range. It is silly to pay the cost of a function call for every single value in every single integer column, even if that function just returns.

ienkovich: On one side, making a range check combined with stats computation is very reasonable and reduces the number of scans. On the other side, our only external data source is Arrow, which uses another decimal format and always requires conversion. So, we can probably make the converter responsible for the check and remove the decimal validator from encoders.

alexbaden:
As expected this validation also happens during arrow parse:
``` 
C++ exception with description "Invalid: In CSV column #0: Error converting '12345678910.12345' to decimal128(10, 4): precision not supported by type." thrown in the test body.
```
so as long as the arrow and HDK types match, decimal validation is actually completely unnecessary. I have updated the PR to remove it entirely. 
I also added my commits to fixup table stats so I could make sure tests pass in lazy mode.

ienkovich (approved):

PR #615, opened 2023-08-03
Introduce simplified string dictionary 

alexbaden:
This PR introduces a drop-in replacement for the `StringDictionary` class using STL containers under the namespace `fast`. The previous string dictionary implementation remains under the namespace `legacy`. The "fast" dictionary has about the same performance of the previous dictionary, but includes options like materialized hashes on by default (materialized hashes give a small performance penalty for small dictionaries in the legacy implementation, because the cost of materializing and storing the hash outweighs the benefit of skipping string comparisons when there are lots of collisions in the hash table). Materialized hashing is essentially free in the fast implementation because we store the hashes directly next to the string ID in the hash table payload, and both are 4 byte values. 

Currently fast dictionary passes all StringDict tests **except** for those involved in dictionary translation. I intend to push the performance boundary of the fast dictionary first, then enable all tests, and finally remove the legacy dictionary and add APIs which should give more performance - e.g. where string ownership remains outside of the dictionary. Also, this PR depends on #613 and #614 which should be merged first.

### DIFF
--- a/omniscidb/StringDictionary/StringDictionary.cpp
+++ b/omniscidb/StringDictionary/StringDictionary.cpp
@@ -50,7 +50,7 @@ namespace {
 
 const int SYSTEM_PAGE_SIZE = omnisci::get_page_size();
 
-uint32_t hash_string(const std::string_view& str) {
+uint32_t hash_string(const std::string_view str) {
   uint32_t str_hash = 1;
   // rely on fact that unsigned overflow is defined and wraps
   for (size_t i = 0; i < str.size(); ++i) {
@@ -77,6 +77,9 @@ struct ThreadInfo {
 }  // namespace
 
 bool g_enable_stringdict_parallel{false};
+
+namespace legacy {
+
 constexpr int32_t StringDictionary::INVALID_STR_ID;
 constexpr size_t StringDictionary::MAX_STRLEN;
 constexpr size_t StringDictionary::MAX_STRCOUNT;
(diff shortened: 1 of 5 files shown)


omniscidb/StringDictionary/StringDictionary.cpp
```diff
@@ -1408,3 +1416,315 @@ size_t StringDictionary::buildDictionaryTranslationMap(
...
+                                          const int64_t generation) const;
+
+template <class T, class String>
+void StringDictionary::getOrAddBulk(const std::vector<String>& string_vec,
+                                    T* output_string_ids) {
+  mapd_lock_guard<mapd_shared_mutex> write_lock(rw_mutex_);
+
+  // compute hashes
+  auto hashes = std::make_unique<uint32_t[]>(string_vec.size());
+  tbb::parallel_for(tbb::blocked_range<size_t>(0, string_vec.size()),
```
ienkovich, line 1508: The legacy version used a grain size to avoid very small tasks. I think it's reasonable to put some limits here. We don't want to create a task per each string.

ienkovich (approved):
Overall looks good! I have a couple of minor threading concerns though.

omniscidb/StringDictionary/StringDictionary.cpp
alexbaden: Thanks for the tips! I experimented with a few different options following [this guide](https://oneapi-src.github.io/oneTBB/main/tbb_userguide/Controlling_Chunking_os.html) and something in the range of 25k seemed to be the best, though not by very much. It's possible taxi is too noisy and we need a dedicated microbenchmark to tune this. I think 25k should be fine for now, though.

Issue #116, opened 2022-11-18
Add shared memory support for L0 path


kurapov-peter:

lmontigny:
Created branch: 116-add-shared-memory-support-for-l0-path

lmontigny:
Moved to pakurapo/smem-2 branch, need to change the **is_l0** flag to true in several place to avoid CPU fallback (and use GPU via L0). Otherwise reported address space casting issue

lmontigny:
Large address space issue related to the trait object.
solved using
```
    if (slot_bytes == sizeof(int32_t)) {
      // return traits.smemPointerType(llvm::Type::getInt32Ty(context));
      return llvm::Type::getInt32PtrTy(context, /*address_space=*/3)
```
generate:
`%dest_slot_adr_0 = bitcast i8 addrspace(3)* %2 to i32 addrspace(3)*`

Current issue:
```
/GpuSharedMemoryTestIntel.INFO
 2023-05-03T04:39:46.972076 I 1772468 0 0 L0Mgr.cpp:75 Discovered 1 driver(s) for L0 platform.
 2023-05-03T04:39:47.003210 I 1772468 0 0 L0Mgr.cpp:221 L0 module build log: error: IGC SPIRV Consumer does    not support the following opcode: 65534
 ^@

```

lmontigny:
L0Mgr issue, it seems to be an invalid opcode as 65535 is not reserved by any vendor:
https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/spir-v.xml#L152

`<ids type="opcode" start="6656" end="65535" comment="Opcode range reservable for future use by vendors"/>`

Wrong data is read as an opcode earlier in the IGC SPIR-V process

lmontigny:
Searching the invalid opcode, see [spir-v disassembly](https://gist.github.com/lmontigny/f4b201f08b412c1dd1915b27b604a03c)

lmontigny:
Added `write_projection_int64` in genx.ll with the casting to addrspacecast(4)
`%6 = bitcast i8 addrspace(4)* %0 to i64 addrspace(4)*`

`write_projection_int32` is required in genx.ll as well.

Now same casting issue is appearing for `agg_sum_double_skip_val`, this function needs to be added in genx.ll as well. (with the casting addrspace(4)).

lmontigny:
Fixed several function and added implementation in genx.ll

Need to fix this one:
```
[...]

PR #618, opened 2023-08-04
Refactor GPU shared memory tests and add L0-specific ones [1/N]

kurapov-peter:
This PR makes most of the code for shared memory tests much more generic. Those parts that are platform specific are renamed.
It also adds an analogous test for L0, though not all the cases work now with shared memory. Those that do are enabled.


kurapov-peter:
The failure is similar to what we have previously seen (https://github.com/intel-ai/hdk/issues/116#issuecomment-1563088825). It does reproduce locally with a docker build on a machine that runs the tests successfully in bare-metal mode. The hypothesis is that the driver mismatch causes the flow to use an older version of spirv consumer. Will try to update the driver first (https://github.com/intel-ai/hdk/pull/622).

kurapov-peter:
The hypothesis proved true, but we'll need yet another driver update as the fix is not yet available in the latest release. It is expected to be published in ~3 weeks. I'm disabling regular runs and will continue with local checks meanwhile.

PR #619, opened 2023-08-04
Assign unique id to Node copy.

### LINKED ISSUE #617: Execution aborted in taxi Q4 on randomly generated taxi data
This bug happens on randomly generated taxi data. Download a data file from here: https://modin-datasets.s3.us-west-2.amazonaws.com/taxi/taxi-random.csv . Get the benchmark code from here: https://github.com/gshimansky/data-science-processing-workload/blob/master/benchmarks/taxi.py
I am using modin 0.23 and pyhdk 0.7 from conda-forge on Linux. When I run taxi benchmark like this I am getting failed assertion somewhere in HDK:
```
$ python taxi.py taxi-random.csv
UserWarning: 'dropna' is temporary disabled due to https://github.com/modin-project/modin/issues/2896
[2023-08-03 19:11:46.326111] [0x00007f8b4e900740] [info]    0 0 RelAlgExecutor.h:233 Check failed: it_ok.second
Aborted
```

ienkovich:
We assume that all nodes in DAG have unique IDs but `Node::deepCopy` method returns a node with the same ID. Fix it by defining a copy ctor for `Node`.

Resolves #617

### DIFF
--- a/omniscidb/IR/Node.cpp
+++ b/omniscidb/IR/Node.cpp
@@ -75,6 +75,12 @@ Node::Node(NodeInputs inputs)
     , context_data_(nullptr)
     , is_nop_(false) {}
 
+Node::Node(const Node& other)
+    : inputs_(other.inputs_)
+    , id_(crt_id_.fetch_add(1))
+    , context_data_(nullptr)
+    , is_nop_(other.is_nop_) {}
+
 void Node::replaceInput(
     NodePtr old_input,
     NodePtr input,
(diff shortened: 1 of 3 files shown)


omniscidb/IR/Node.cpp
```diff
@@ -75,6 +75,12 @@ Node::Node(NodeInputs inputs)
     , context_data_(nullptr)
     , is_nop_(false) {}
 
+Node::Node(const Node& other)
+    : inputs_(other.inputs_)
+    , id_(crt_id_.fetch_add(1))
+    , context_data_(nullptr)
```
kurapov-peter, line 81: So this is basically a shallow copy. Do we need it at all? I.e. would it be reasonable to have the copy constructor do a deep copy?

kurapov-peter (approved):

PR #622, opened 2023-08-08
Move to ubuntu 22.04 in docker files and update compute-runtime version

kurapov-peter:
p.s. why do we have Dockerfile.clang that is identical to Dockerfile.cuda?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,13 +24,8 @@ RUN apt-get update && apt-get install -y \
 
 # Toolchains
 RUN apt-get update && apt-get install -y \
-    gcc-9 \
-    g++-9 \
     openjdk-11-jdk \
-    -- && \
-    update-alternatives \
-        --install /usr/bin/gcc gcc /usr/bin/gcc-9 800 \
-        --slave /usr/bin/g++ g++ /usr/bin/g++-9
+    --
 
 # LLVM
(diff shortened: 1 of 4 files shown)


kurapov-peter:
The cause of the installation failure is that `cuda-drivers-525` fails to build the module (similar to https://forums.developer.nvidia.com/t/510-68-02-driver-breaks-installation-for-newer-linux-kernel-5-17-8-on-ubuntu-20-04/214531):
```
gcc: error: unrecognized command line option '-mharden-sls=all'
make[2]: *** [scripts/Makefile.build:297: /var/lib/dkms/nvidia/525.125.06/build/nvidia/nv.o] Error 1
```

Autoconf reports a compiler version mismatch:
```
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: gcc (Ubuntu 11.3.0-1ubuntu1~22.04.1) 11.3.0
  You are using:           gcc (Ubuntu 9.5.0-1ubuntu1~22.04) 9.5.0
```

This is caused by our docker file to `update-alternative` and set gcc default version to 9.5. Setting the default version back to 11 (even though the installed one is 11.4) solved the problem locally.

kurapov-peter:
Now it is `fatal error: 'cstdint' file not found`. We already saw this some time ago when we tried to update gcc. The stupid solution was to install g++-12. I tried installing libstdc++ separately and clang finds it after installation, but `cstdint` is still missing.

leshikus (approved):
LGTM

Issue #574, opened 2023-07-06
[Perf][Bench] Join is slow on big tables.


Devjiu:
On size of tables ~10^8 join performance is very low, compared with duckdb. 

On join by single int column for 2 tables 10^8 size takes about 12 s. 
```js
12888ms total duration for executeRelAlgQuery
        12886ms start(1ms) executeWorkUnit RelAlgExecutor.cpp:1389
          569ms start(1ms) compileWorkUnit NativeCodegen.cpp:1403
          5394ms start(571ms) fetchChunks Execute.cpp:3090
          67ms start(5966ms) executePlan Execute.cpp:3462
          9ms start(6487ms) collectAllDeviceResults Execute.cpp:2550
          70ms start(6497ms) compileWorkUnit NativeCodegen.cpp:1403
          3071ms start(6568ms) fetchChunks Execute.cpp:3090
          527ms start(9640ms) executePlan Execute.cpp:3462
``` 

The biggest performance drop is in fetchChunks. There are also 2 subqueries, the first is `COUNT(*)` to collect metadata.

Devjiu:
Possible points/hints to increase performance:

1. There are too much copying of data in getAllTableColumnFragments.
```js
    585ms start(597ms) getAllTableColumnFragments ColumnFetcher.cpp:243
        68ms start(597ms) getOneTableColumnFragment ColumnFetcher.cpp:176
        58ms start(666ms) ColumnarResults ColumnarResults.cpp:124
        69ms start(724ms) getOneTableColumnFragment ColumnFetcher.cpp:176
        57ms start(793ms) ColumnarResults ColumnarResults.cpp:124
        69ms start(851ms) getOneTableColumnFragment ColumnFetcher.cpp:176
        57ms start(921ms) ColumnarResults ColumnarResults.cpp:124
        8ms start(979ms) getOneTableColumnFragment ColumnFetcher.cpp:176
        7ms start(987ms) ColumnarResults ColumnarResults.cpp:124
        187ms start(995ms) mergeResults ColumnarResults.cpp:139
```
Looks like there are 3 chunk copies. 
- getChunkBuffer - should be copy from Arrow and aimed to create chunks
  - createZeroCopyBuffer - is fast and fine
  - if not we are going to persistent DataMgr, than to ArrowStorage::fetchBuffer <- _this is slow._
- mergeResults is too slow
1st copy - is copy from specific fragment to required format (arrow), 
[...]

PR #623, opened 2023-08-08
[Join] Zero-copy storage column check.

Devjiu:
This commit adds check for valid pointer to column buffer to skip copying. Most effective with enabled `enable-non-lazy-data-import` option. Checks number of chunks in storage.

Partially resolves: #574

### DIFF
--- a/omniscidb/ArrowStorage/ArrowStorage.cpp
+++ b/omniscidb/ArrowStorage/ArrowStorage.cpp
@@ -187,6 +187,51 @@ std::unique_ptr<AbstractDataToken> ArrowStorage::getZeroCopyBufferMemory(
   return nullptr;
 }
 
+std::unique_ptr<AbstractDataToken> ArrowStorage::getZeroCopyColumnData(
+    const ColumnRef& col_ref) {
+  mapd_shared_lock<mapd_shared_mutex> data_lock(data_mutex_);
+  CHECK_EQ(col_ref.db_id, db_id_);
+  CHECK_EQ(tables_.count(col_ref.table_id), (size_t)1);
+  auto& table = *tables_.at(col_ref.table_id);
+  mapd_shared_lock<mapd_shared_mutex> table_lock(table.mutex);
+  data_lock.unlock();
+
+  auto col_type = getColumnInfo(col_ref.db_id, col_ref.table_id, col_ref.column_id)->type;
+
+  if (col_type->isExtDictionary()) {
+    auto dict_id = col_type->as<hdk::ir::ExtDictionaryType>()->dictId();
+    auto dict_descriptor = getDictMetadata(
+        dict_id);  // this will force materialize the dictionary. it is thread safe
+    CHECK(dict_descriptor);
+  }
+
+  if (!col_type->isVarLen()) {
+    size_t col_idx = columnIndex(col_ref.column_id);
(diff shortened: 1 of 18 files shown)


omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -187,6 +187,53 @@ std::unique_ptr<AbstractDataToken> ArrowStorage::getZeroCopyBufferMemory(
   return nullptr;
 }
 
+std::unique_ptr<AbstractDataToken> ArrowStorage::getZeroCopyColumnData(
+    const ColumnRef& col_ref) {
```
Devjiu, line 191: Almost copy-paste of `getZeroCopyBufferMemory`. Should we unite this methods, or add private utility method?

omniscidb/ResultSetRegistry/ResultSetRegistry.cpp
```diff
@@ -436,6 +436,12 @@ ResultSetRegistry::getZeroCopyBufferMemory(const ChunkKey& key, size_t num_bytes
              : nullptr;
 }
 
+std::unique_ptr<AbstractDataToken> ResultSetRegistry::getZeroCopyColumnData(
+    const ColumnRef& col_ref) {
+  // UNREACHABLE();
```
Devjiu, line 441: not sure how to handle this case correctly, should we support it?

omniscidb/ArrowStorage/ArrowStorage.cpp
Devjiu: Marking it as temporary solution, should be removed in further prs.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -187,6 +187,51 @@ std::unique_ptr<AbstractDataToken> ArrowStorage::getZeroCopyBufferMemory(
...
+    }
+    size_t elem_size = col_type->size();
+    const auto* fixed_type =
+        dynamic_cast<const arrow::FixedWidthType*>(table.col_data[col_idx]->type().get());
+    CHECK(fixed_type) << table.col_data[col_idx]->type()->ToString() << " (table "
+                      << col_ref.table_id << ", column " << col_idx << ")";
+    size_t arrow_elem_size = fixed_type->bit_width() / 8;
+    size_t elems = elem_size / arrow_elem_size;
+    CHECK_GT(elems, (size_t)0);
+    auto data_to_fetch = table.col_data[col_idx];
```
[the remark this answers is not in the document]
Devjiu: Fixed PTAL

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -203,7 +248,7 @@ void ArrowStorage::fetchFixedLenData(const TableData& table,
   // to get a proper slice.
   size_t elems = elem_size / arrow_elem_size;
   CHECK_GT(elems, (size_t)0);
-  auto data_to_fetch =
+  const auto& data_to_fetch =
```
ienkovich, line 251: `Slice` creates a new `shared_ptr`, therefore you cannot use reference here and all other similar places.

ienkovich (approved):

PR #624, opened 2023-08-09
Add support for LLVM 16

alexbaden:
Starting with the build changes, lots of assertions due to opaque pointer usage. It still compiles, but taking the type is now a runtime error (which seems like an annoying way to do it).

### DIFF
--- a/omniscidb/QueryEngine/QueryTemplateGenerator.cpp
+++ b/omniscidb/QueryEngine/QueryTemplateGenerator.cpp
@@ -601,12 +601,12 @@ class GroupByQueryTemplateGenerator : public QueryTemplateGenerator {
     if (query_mem_desc.hasVarlenOutput()) {
       // make the varlen buffer the _first_ 8 byte value in the group by buffers double
       // ptr, and offset the group by buffers index by 8 bytes
-      // TODO(llvm16): should this be 32?
+      // TODO(llvm16): address space
       auto varlen_output_buffer_gep = llvm::GetElementPtrInst::Create(
-          get_int_type(64, mod->getContext()),
+          llvm::PointerType::get(mod->getContext(), 0),
           output_buffers,
           llvm::ConstantInt::get(llvm::Type::getInt32Ty(mod->getContext()), 0),
-          "",
+          "col_buffer_ptr",
           bb_entry);
       row_func_call_args->varlen_output_buffer =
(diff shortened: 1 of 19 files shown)


alexbaden:
The great pointer type purge begins - so far, not awful, though there are a few places where there might be some confusion. We have 77 instances of `getPointerElementType()` left.

omniscidb/QueryEngine/QueryTemplateGenerator.cpp
```diff
@@ -561,6 +552,7 @@ class GroupByQueryTemplateGenerator : public QueryTemplateGenerator {
     auto crt_matched_uncasted_ptr =
         new llvm::AllocaInst(i32_type, 0, "crt_matched", bb_entry);
     if (crt_matched_uncasted_ptr->getType() != pi32_type) {
+      // can this ever happen?
```
kurapov-peter, line 555: I think it always happens if recall correctly. Alloca returns a pointer type with the default address space, i.e. zero.

omniscidb/QueryEngine/QueryTemplateGenerator.cpp
```diff
@@ -1019,26 +1011,25 @@ class NonGroupedQueryTemplateGenerator : public QueryTemplateGenerator {
...
+          // TODO(adb): (1) need to set address space on this opaque pointer properly
+          //            (2) need to make sure this sequence is correct w/r/t opague
+          //            pointer types
           auto out_gep = llvm::GetElementPtrInst::CreateInBounds(
-              output_buffers->getType()->getPointerElementType(),
-              output_buffers,
-              col_idx,
+              get_int_type(64, mod->getContext()), output_buffers, col_idx, "", bb_exit);
+          auto col_buffer = new llvm::LoadInst(
+              llvm::PointerType::get(mod->getContext(), /*AddressSpace=*/0),
```
kurapov-peter, line 1020: Input column buffer, as well as output buffer pointers, have global (1) address space at the topmost level. The wrapper casts them to generic (4, this matches our local type in the codegen traits; I should rename it to "generic" for clarity) and calls the query stub. GEP creation does not require explicit address space mentioning though, you can use `getType()`, here's an example: https://github.com/kurapov-peter/L0Snippets/blob/main/common/helpers.cpp#L46

alexbaden:
Superseded by #658 .

Issue #572, opened 2023-07-06
Allow different fragment sizes for heterogeneous execution mode


kurapov-peter:
Currently, the fragment size is a constant for whether we run on a CPU or GPU. In heterogeneous execution mode this results in suboptimal resource utilization. Allowing different fragment sizes for different targets can help.

PR #628, opened 2023-08-10
Refragmenting table

akroviakov:
This PR introduces a table refragmentation functionality, which allows to avoid doing an import of the same table just for a new fragment size.
It is relatively cheaper to refragment a table (as opposed to re-inserting it), since we don't have to modify the physical layout/location of the data in memory. The price we pay for refragmenting is recalculating the metadata (basically fragment's offset, row_count + small materialized aggregates per fragment) in order to change the logical view of the table. 

Addresses #572 .

Spends quite some time in `ArrowStorage::computeStats()`, the code for `FixedLengthEncoder::updateStatsEncoded()` appears to be scalar, possible performance improvement opportunity for calculating min/max/detecting nulls using SIMD.

### DIFF
--- a/omniscidb/ArrowStorage/ArrowStorage.cpp
+++ b/omniscidb/ArrowStorage/ArrowStorage.cpp
@@ -630,24 +630,79 @@ TableInfoPtr ArrowStorage::importArrowTable(std::shared_ptr<arrow::Table> at,
   return importArrowTable(at, table_name, columns, options);
 }
 
-void ArrowStorage::refragmentTable(const std::string& table_name,
-                                   const size_t new_frag_size) {
-  auto tinfo = getTableInfo(db_id_, table_name);
-  if (!tinfo) {
-    throw std::runtime_error("Unknown table: "s + table_name);
-  }
+TableInfoPtr ArrowStorage::createRefragmentedView(const std::string& table_name,
+                                                  const std::string& new_table_name,
+                                                  const size_t new_frag_size) {
   if (!new_frag_size) {
     throw std::runtime_error("Cannot refragment to fragment size 0");
   }
-  mapd_shared_lock<mapd_shared_mutex> data_lock(data_mutex_);
-  auto& table = *tables_.at(tinfo->table_id);
-  mapd_unique_lock<mapd_shared_mutex> table_lock(table.mutex);
+
+  if (table_name.empty()) {
+    throw std::runtime_error("Cannot create table with empty name");
+  }
+  TableInfoPtr res;
+  int new_table_id;
+  TableData original_table_data;
+  mapd_unique_lock<mapd_shared_mutex> data_lock(data_mutex_);
+  {
+    mapd_unique_lock<mapd_shared_mutex> dict_lock(dict_mutex_);
+    mapd_unique_lock<mapd_shared_mutex> schema_lock(schema_mutex_);
+    if (getTableInfoNoLock(db_id_, new_table_name)) {
+      throw std::runtime_error("Table with name '"s + new_table_name +
+                               "' already exists"s);
+    }
+
+    auto tinfo = getTableInfoNoLock(db_id_, table_name);
+    if (!tinfo) {
+      throw std::runtime_error("Unknown table: "s + table_name);
+    }
+    auto& table = *tables_.at(tinfo->table_id);
+    original_table_data.schema = table.schema;
+    original_table_data.col_data = table.col_data;
+    original_table_data.row_count = table.row_count;
+    original_table_data.table_stats = table.table_stats;
+    new_table_id = next_table_id_++;
+    res = addTableInfo(db_id_, new_table_id, new_table_name, false, 0, 0);
+    auto col_info_list = listColumnsNoLock(db_id_, tinfo->table_id);
+    for (size_t col_idx = 0; col_idx < col_info_list.size() - 1; col_idx++) {
+      const auto col_info = col_info_list[col_idx];
+      CHECK(col_info);
+      auto col_type = col_info->type;
+      auto col_arr = table.col_data[col_idx];
+      addColumnInfo(
+          db_id_, new_table_id, columnId(col_idx), col_info->name, col_type, false);
+      if (col_type->isExtDictionary() && col_arr->type()->id() == arrow::Type::STRING) {
+        auto dict_data =
+            dicts_.at(dynamic_cast<const hdk::ir::ExtDictionaryType*>(col_type)->dictId())
+                .get();
+        CHECK(dict_data);
+        dict_data->addTableColumnPair(new_table_id, col_idx);
+      }
+    }
+  }
+  auto [iter, inserted] = tables_.emplace(new_table_id, std::make_unique<TableData>());
+  CHECK(inserted);
+
+  auto& new_table = *iter->second;
+  mapd_unique_lock<mapd_shared_mutex> table_lock(new_table.mutex);
   data_lock.unlock();
-  if (new_frag_size == table.fragment_size) {
-    return;
+  new_table.schema = original_table_data.schema;
+  new_table.col_data = original_table_data.col_data;
+  new_table.row_count = original_table_data.row_count;
+  new_table.table_stats = original_table_data.table_stats;
+
+  refragmentTable(new_table, new_table_id, new_frag_size);
+  return res;
+}
+
+void ArrowStorage::refragmentTable(TableData& table,
+                                   const int table_id,
+                                   const size_t new_frag_size) {
+  if (!new_frag_size) {
+    throw std::runtime_error("Cannot refragment to fragment size 0");
   }
   const size_t new_frag_count =
-      table.row_count / new_frag_size + ((table.row_count % new_frag_size) != 0);
+      std::ceil(static_cast<float>(table.row_count) / new_frag_size);
   std::vector<DataFragment> new_fragments(new_frag_count);
 
   for (size_t new_frag_idx = 0; new_frag_idx < new_frag_count - 1; new_frag_idx++) {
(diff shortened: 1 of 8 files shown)


ienkovich:
It is not enough to simply refragment a table inside the storage. Assigning the new fragment size means you change data assigned to already used chunk IDs. This means all data cached in buffer managers becomes irrelevant but you would still use it in the following queries.

There is other data that is referenced by chunk ID and can be cached, e. g. chunk stats.

So, refragmenting needs additional cleanups but it's not clear how to do it all atomically.

Did you consider creating a new table as a result of refragmentation? Same price, no conflicts.

akroviakov:
@ienkovich Do we update anything above the storage layer when we append to a table and the last fragment wasn't full and now it starts to span arrow's chunks boundaries (and so we update its metadata)?  https://github.com/intel-ai/hdk/blob/ad924b50a2e909a98efba9a6363f9b5415742bf6/omniscidb/ArrowStorage/ArrowStorage.cpp#L876

ienkovich:
> @ienkovich Do we update anything above the storage layer when we append to a table and the last fragment wasn't full and now it starts to span arrow's chunks boundaries (and so we update its metadata)?

Appended data was originally supported in buffer managers. The size of the chunk grows and buffer managers ask a storage for new data. If you reduce the size of a chunk then they would simply use the old data.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -630,6 +630,177 @@ TableInfoPtr ArrowStorage::importArrowTable(std::shared_ptr<arrow::Table> at,
...
+  }
+  if (!new_frag_size) {
+    throw std::runtime_error("Cannot refragment to fragment size 0");
+  }
+  mapd_shared_lock<mapd_shared_mutex> data_lock(data_mutex_);
+  auto& table = *tables_.at(tinfo->table_id);
+  mapd_unique_lock<mapd_shared_mutex> table_lock(table.mutex);
+  data_lock.unlock();
+  const size_t new_frag_count =
+      table.row_count / new_frag_size + ((table.row_count % new_frag_size) != 0);
```
Devjiu, line 705: Maybe `std::ceil` ? It will be more clear.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -630,6 +630,177 @@ TableInfoPtr ArrowStorage::importArrowTable(std::shared_ptr<arrow::Table> at,
...
+    for (auto col_idx = range.begin(); col_idx != range.end(); col_idx++) {
+      auto col_info = getColumnInfo(db_id_, tinfo->table_id, columnId(col_idx));
+      auto col_type = col_info->type;
+      auto col_arr = table.col_data[col_idx];
+      auto elem_type =
+          col_type->isArray()
+              ? dynamic_cast<const hdk::ir::ArrayBaseType*>(col_type)->elemType()
+              : col_type;
+      bool compute_stats = !col_type->isString();
+      if (compute_stats) {
```
Devjiu, line 748: I think it's better to move this condition out of cycle. Earlier you already checked which frag_ids that have String type.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -1070,26 +1241,31 @@ void ArrowStorage::dropTable(int table_id, bool throw_if_not_exist) {
...
     }
   }
 
   SimpleSchemaProvider::dropTable(db_id_, table_id);
   // TODO: clean-up shared dictionaries without a full scan of
   // existing columns.
-  for (auto& pr : column_infos_) {
-    if (pr.second->type->isExtDictionary()) {
-      dicts_to_remove.erase(pr.second->type->as<hdk::ir::ExtDictionaryType>()->dictId());
+  if (!is_view) {
```
Devjiu, line 1258: simplify it please.
```c++ 
if (is_view) {
  return;
}
```

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -1070,26 +1241,31 @@ void ArrowStorage::dropTable(int table_id, bool throw_if_not_exist) {
   std::unique_ptr<TableData> table = std::move(tables_.at(table_id));
   mapd_unique_lock<mapd_shared_mutex> table_lock(table->mutex);
   tables_.erase(table_id);
-
   std::unordered_set<int> dicts_to_remove;
-  auto col_infos = listColumnsNoLock(db_id_, table_id);
-  for (auto& col_info : col_infos) {
-    if (col_info->type->isExtDictionary()) {
-      dicts_to_remove.insert(col_info->type->as<hdk::ir::ExtDictionaryType>()->dictId());
+  if (!is_view) {
```
Devjiu, line 1245: same simplification here, please.

omniscidb/ArrowStorage/ArrowStorage.h
```diff
@@ -85,6 +85,13 @@ class ArrowStorage : public SimpleSchemaProvider, public AbstractDataProvider {
   TableInfoPtr importArrowTable(std::shared_ptr<arrow::Table> at,
                                 const std::string& table_name,
                                 const TableOptions& options = TableOptions());
+  TableInfoPtr getRefragmentedView(const std::string& table_name,
+                                   const std::string& new_table_name,
+                                   const size_t new_frag_size);
+  TableInfoPtr createRefragmentedView(const std::string& table_name,
+                                      const std::string& new_table_name,
+                                      const size_t new_frag_size);
+  void refragmentTable(const std::string& table_name, const size_t new_frag_size);
```
Devjiu, line 94: Do we need this methods in public? As far as I see `getRefragmentedView` uses both this  methods, maybe only this method is enough?

ienkovich (changes requested):
The general implementation looks fine, but there are some synchronization issues. To correctly create new table you need:
1. Get data unique lock
2. Get schema unique lock
3. Get dict unique lock
4. Make all checks and add a new table + modify dict data
5. Release schema and dict locks
6. Create new table data
7. Get a table unique lock
8. Release the data lock
9. Make the rest of the manipulations with the table
10. Release the table lock

It's hard to make it using all three methods taking the table name as a parameter. I suggest you have `createRefragmentedView` as the entry point (better name than `getRefragmentedView` because it is not a simple getter), do steps 1-8 there, and then call `refragmentTable` passing TableData reference as a param. Then you should be thread-safe.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -630,6 +630,176 @@ TableInfoPtr ArrowStorage::importArrowTable(std::shared_ptr<arrow::Table> at,
...
 }
 
+TableInfoPtr ArrowStorage::createRefragmentedView(const std::string& table_name,
+                                                  const std::string& new_table_name,
+                                                  const size_t new_frag_size) {
+  if (!new_frag_size) {
+    throw std::runtime_error("Cannot refragment to fragment size 0");
+  }
+
+  if (table_name.empty()) {
```
ienkovich, line 640: `new_table_name`?

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -630,6 +630,176 @@ TableInfoPtr ArrowStorage::importArrowTable(std::shared_ptr<arrow::Table> at,
...
+  new_table.table_stats = original_table_data.table_stats;
+
+  refragmentTable(new_table, new_table_id, new_frag_size);
+  return res;
+}
+
+void ArrowStorage::refragmentTable(TableData& table,
+                                   const int table_id,
+                                   const size_t new_frag_size) {
+  if (!new_frag_size) {
```
ienkovich, line 701: Duplicated check that makes an impression that we create a new table and then leave it with uninitialized fragments in case of zero fragment size. I'd use `CHECK` instead here.

omniscidb/ArrowStorage/ArrowStorage.cpp
```diff
@@ -630,6 +630,176 @@ TableInfoPtr ArrowStorage::importArrowTable(std::shared_ptr<arrow::Table> at,
...
+    throw std::runtime_error("Cannot refragment to fragment size 0");
+  }
+  const size_t new_frag_count =
+      std::ceil(static_cast<float>(table.row_count) / new_frag_size);
```
ienkovich, line 705: Float precision is not good enough here. If you have 1'000'000'001 rows and fragment size is 10'000'000 then your result here would be 100 fragments, not 101. Use integer arithmetics to avoid precision issues.

python/pyhdk/_storage.pxd
```diff
@@ -131,7 +131,7 @@ cdef extern from "omniscidb/ArrowStorage/ArrowStorage.h":
     CArrowStorage(int, string, int, shared_ptr[CConfig]) except +;
 
     CTableInfoPtr createTable(const string&, const vector[CColumnDescription]&, const CTableOptions&) except +
-
+    CTableInfoPtr createRefragmentedView(const string& , const string&, const size_t) except +
```
ienkovich, line 134: Our main API access point is `pyhdk.HDK` class. Do you use it in your demo? I'd expect usage scenario to be something like:
```
ht1 = hdk.import_csv(file, fragment_size=10)
...
ht2 = ht1.refragmented_view(fragment_size=20)
...
```

ienkovich (changes requested):

python/pyhdk/_storage.pxd
akroviakov: Demo will be redone as a separate PR.

ienkovich (approved):
Looks good!

alexbaden:
It would be nice to have a googletest / c++ test for this - also I am wondering what happens if you enable lazy dictionary materialization then re-fragment. Could we add something to `ArrowStorageTest.cpp`?

PR #629, opened 2023-08-10
Fix arrays fetch in ResultSetRegistry.

### LINKED ISSUE #627: QueryBuilderTest.FixedArrayInRes produces incorrect results with allow_lazy_fetch=true
Setting the default `true` value for `CompilationOptions::allow_lazy_fetch` makes `QueryBuilderTest` produce incorrect results two cases.
`QueryBuilderTest.VarlenArrayInRes` and `QueryBuilderTest.FixedArrayInRes` are affected.

```
[ RUN      ] QueryBuilderTest.FixedArrayInRes
/localdisk/pakurapo/sources/hdk/omniscidb/Tests/ArrowTestHelpers.h:189: Failure
Value of: chunk->IsNull(j)
  Actual: false
Expected: true
/localdisk/pakurapo/sources/hdk/omniscidb/Tests/ArrowTestHelpers.h:189: Failure
Value of: chunk->IsNull(j)
  Actual: false
Expected: true
/localdisk/pakurapo/sources/hdk/omniscidb/Tests/ArrowTestHelpers.h:189: Failure
Value of: chunk->IsNull(j)
  Actual: false
Expected: true
/localdisk/pakurapo/sources/hdk/omniscidb/Tests/ArrowTestHelpers.h:189: Failure
Value of: chunk->IsNull(j)
  Actual: false
Expected: true
```

ienkovich:
Currently, NULL value for fixed-length arrays in TargetValue can be expressed in two different ways - nullopt and zero-length array. Only arrow converter supports both cases, fetch and contentToString don't cover both. This patch forces nullopt to be always used.

Also, this patch fixes the fetch of varlen arrays with lazy fetch.

Resolves #627

### DIFF
--- a/omniscidb/ResultSet/ArrowResultSetConverter.cpp
+++ b/omniscidb/ResultSet/ArrowResultSetConverter.cpp
@@ -1057,29 +1057,17 @@ size_t convert_rowwise(
         int32_t arr_size = 0;
         if (*array_value) {
           arr_size = static_cast<int32_t>((*array_value)->size());
-          // Fixed size array NULL value can be expressed as a zero-length vector.
-          // We should add a proper number of empty values for such case.
-          if (column.col_type->isFixedLenArray() && arr_size != num_elems) {
-            CHECK_EQ(arr_size, 0);
-            append_empty_values_and_validity(elem_type,
-                                             num_elems,
+          for (auto& elem_value : **array_value) {
+            append_scalar_value_and_validity(elem_value,
+                                             elem_type,
                                              device_type,
                                              value_seg[j],
                                              null_bitmap_seg[j],
                                              max_size);
-          } else {
-            for (auto& elem_value : **array_value) {
-              append_scalar_value_and_validity(elem_value,
-                                               elem_type,
-                                               device_type,
-                                               value_seg[j],
-                                               null_bitmap_seg[j],
-                                               max_size);
-            }
           }
         } else if (column.col_type->isFixedLenArray()) {
-          // Assume this is also a possible option for NULL fixed size array
-          // and handle appropriately.
+          // We got NULL fixed-length array and need to add a proper number of empty
+          // values.
           append_empty_values_and_validity(elem_type,
                                            num_elems,
                                            device_type,
@@ -1105,8 +1093,7 @@ size_t convert_rowwise(
             validity = std::make_shared<std::vector<uint8_t>>();
             validity->reserve(local_entry_count);
           }
-          int8_t is_valid =
-              *array_value && (!column.col_type->isFixedLenArray() || arr_size != 0);
+          int8_t is_valid = (bool)*array_value;
           validity->push_back(is_valid);
         }
       } else {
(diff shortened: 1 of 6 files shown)


omniscidb/ResultSet/ArrowResultSetConverter.cpp
```diff
@@ -1057,29 +1057,17 @@ size_t convert_rowwise(
         int32_t arr_size = 0;
         if (*array_value) {
           arr_size = static_cast<int32_t>((*array_value)->size());
-          // Fixed size array NULL value can be expressed as a zero-length vector.
-          // We should add a proper number of empty values for such case.
-          if (column.col_type->isFixedLenArray() && arr_size != num_elems) {
-            CHECK_EQ(arr_size, 0);
-            append_empty_values_and_validity(elem_type,
-                                             num_elems,
+          for (auto& elem_value : **array_value) {
```
kurapov-peter, line 1060: Can this dereference fail?

omniscidb/ResultSetRegistry/ColumnarResults.cpp
```diff
@@ -458,10 +458,11 @@ inline void ColumnarResults::writeBackCell(const TargetValue& col_val,
       for (auto& elem_val : **arr_col_val) {
         write_scalar(&elem_val, offset++, elem_type);
       }
+    } else if (type->isFixedLenArray()) {
       // Put NULL sentinel value for fixed length array
-      if (type->isFixedLenArray() && (*arr_col_val)->size() == 0) {
```
kurapov-peter, line 462: Are `(*arr_col_val)->size()` and `(bool)(*arr_col_val)` the same?

omniscidb/ResultSet/ResultSet.cpp
```diff
@@ -1055,7 +1055,7 @@ size_t ResultSet::computeVarLenOffsets(size_t col_idx, int32_t* offsets) const {
   }
 
   // Translate varlen value to its length. Return -1 for NULLs.
-  auto slot_val_to_length = [this, lazy_fetch, col_idx, type, target_info](
+  auto slot_val_to_length = [this, lazy_fetch, col_idx, type, target_info, arr_elem_size](
```
kurapov-peter, line 1058: I'd change the name since you've modified it's meaning.

omniscidb/ResultSet/ArrowResultSetConverter.cpp
ienkovich: No, this is a dereference of `boost::optional` with the preceding check that it is not empty.

omniscidb/ResultSetRegistry/ColumnarResults.cpp
ienkovich: `*arr_col_val` is `boost::optional`, so `bool` operator checks it is not empty and `operator->` accesses stored value.

omniscidb/ResultSet/ResultSet.cpp
ienkovich: I think the name still correctly describes the method's meaning. It gets a slot value and returns the corresponding array length in bytes.

kurapov-peter (approved):

PR #634, opened 2023-08-16
Remove implicit cast to size_t

kurapov-peter:


ienkovich:
Looks like this reveals a bug of `get_bit_width` usage with varlen array types.

omniscidb/QueryEngine/OutputBufferInitialization.cpp
```diff
@@ -116,7 +116,7 @@ int64_t get_agg_initial_val(hdk::ir::AggType agg,
   CHECK(!(type->isString() || type->isExtDictionary()) ||
         (agg == hdk::ir::AggType::kSingleValue || agg == hdk::ir::AggType::kSample));
   const auto byte_width =
-      enable_compaction
+      enable_compaction && type->size() > 0
```
alexbaden, line 119: I am surprised this didn't cause errors earlier - is it possible for `enable_compaction` to ever be true here?

alexbaden: Varlen aggregate should also work for `SAMPLE(varlen_type)`, though in that case the aggregation was a hack (we would just place the rowid of the chosen row in the aggregate slot)

ienkovich (approved):

PR #635, opened 2023-08-16
Refactor GPU shared memory tests and add L0-specific ones [2/N]

kurapov-peter:
This patch moves most of the aggregation function implementations to the cpp module. It also applies some fixes that are required since we don't have the automatic zero initialization of the memory. I tried to be as close to cuda implementations as possible to keep it simple and generalize later.


kurapov-peter:
#637

PR #638, opened 2023-08-17
Update conda build env to openjdk 20

alexbaden:


alexbaden:
@leshikus can you take a look at this? We want to update to openjdk 20 in conda for developers and public builds per @vlad-penkin . Older openjdk is fine on the build containers as long as it is only for CI.

PR #641, opened 2023-08-18
Support TopK aggregate in Calcite

AndreyPavlenko:


ienkovich (approved):
OK if it works for Modin.

PR #642, opened 2023-08-18
Disable IR file metadata by default

alexbaden:
I find that this clutters the IR more than it helps. Does anyone use it?


ienkovich (approved):
It's reasonable to have a debug feature disabled by default. No trouble to enable it if you need it.

PR #643, opened 2023-08-18
Add ENABLE_SQL build flag.

ienkovich:
This gives us a build with no SQL and no Java dependency. Still should work with Modin.


alexbaden (approved):
Nice!

ienkovich:
It appears some functions become unavailable in this build. E.g. POWER function. I'll work on it

PR #645, opened 2023-08-22
PVC CI test

kurapov-peter:


pbchekin (approved):
LGTM. Looks like the correct runner was used, successful execution.
I cannot say, however, looking to the logs if GPU was actually used or not.

Devjiu (approved):

PR #650, opened 2023-08-24
Introduce aggregate finalization.

ienkovich:
This patch introduces finalization for aggregates that can be done after reduction. This step can be used to materialize aggregates that do not have the final result computed during reduction (AVG, APPROX_QUANTILE, TOP_K, QUANTILE). For now, I only finalize QUANTILE aggregate as the most time-consuming one.

Currently, actual quantile finalization happens when we try to fetch quantile value the first time and it actually happens when we put ResultSet to the ResultSetRegistry (rowCount iterates through all rows and materializes all values, which is another inefficiency to fix someday). rowCount doesn't expect iteration through rows to be so time-consuming and can choose single-threaded mode in a case when iteration is quite long (>1sec.). Finalization here helps by adding parallelism.

On H2O GroupBy Q6 this patch gives a significant boost speeding up finalization by >10x (due to multithreading) and overall query execution by ~2x for medium dataset.

### DIFF
--- a/omniscidb/ResultSet/ResultSet.cpp
+++ b/omniscidb/ResultSet/ResultSet.cpp
@@ -25,6 +25,7 @@
 
 #include "DataMgr/Allocators/GpuAllocator.h"
 #include "DataMgr/BufferMgr/BufferMgr.h"
+#include "ResultSet/QuantileAccessors.h"
 #include "ResultSet/ResultSet.h"
 #include "ResultSet/TopKAggAccessors.h"
 #include "Shared/InlineNullValues.h"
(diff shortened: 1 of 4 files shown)


omniscidb/ResultSet/ResultSet.cpp
```diff
@@ -1245,6 +1246,118 @@ void ResultSet::initializeStorage() const {
...
+    while (iter.global_entry_idx_valid_) {
+      const auto storage_lookup_result = findStorage(iter.global_entry_idx_);
+      finalize(storage_lookup_result.storage_ptr,
+               storage_lookup_result.fixedup_entry_idx);
+      ++iter;
+      ++row_count;
+    }
+  } else {
+    row_count = tbb::parallel_reduce(
+        tbb::blocked_range((size_t)0, appended_storage_.size() + 1),
```
kurapov-peter, line 1329: Why +1?

omniscidb/ResultSet/ResultSet.cpp
```diff
@@ -1245,6 +1246,118 @@ void ResultSet::initializeStorage() const {
...
+  }
+
+  // This function is going to be called for each non-empty entry.
+  auto finalize = [&](const ResultSetStorage* storage, size_t entry_idx) {
+    for (auto& slot : quantile_slots_to_finalize) {
+      const int8_t* slot_ptr = nullptr;
+      if (query_mem_desc_.didOutputColumnar()) {
+        auto col_ptr =
+            storage->buff_ + storage->query_mem_desc_.getColOffInBytes(std::get<1>(slot));
+        slot_ptr = col_ptr + 8 * entry_idx;
```
kurapov-peter, line 1293: What does 8 stand for?

omniscidb/ResultSet/ResultSet.cpp
```diff
@@ -1245,6 +1246,118 @@ void ResultSet::initializeStorage() const {
...
+
+  // Finalization is usually called as a part of aggregation reduction.
+  // In this case we have a group-by buffer with unknown number of valid
+  // rows. So we can combine finalization with row count computation to
+  // avoid additional data scan later.
+  size_t row_count = 0;
+  // Quantile finalization can be quite costly, so do it in parallel if all
+  // entries are included into the result and we have enough entries.
+  // Otherwise use single-threaded iteration.
+  constexpr size_t min_task_size = 100;
```
kurapov-peter, line 1314: Would it be better to set it to a power of 2?

kurapov-peter (approved):

omniscidb/ResultSet/ResultSet.cpp
ienkovich: Because there is also `storage_`. For some reason, we don't store them all in a single vector.

omniscidb/ResultSet/ResultSet.cpp
ienkovich: The slot size that was checked earlier to be 8.

omniscidb/ResultSet/ResultSet.cpp
ienkovich: there is no benefit in using power of 2 because we don't use it for division, memory allocations, etc.

PR #653, opened 2023-08-25
Improve quantiles allocation.

ienkovich:
Here I stop using `unique_ptr` to allocate and store each individual Quantile object. Instead, I allocate memory from the arena for a bunch of objects and then initialize them manually. So we have only a single lock and allocation per quantile column per allocated result set (similar to what we have for TopK). This dramatically improves the result set initialization time. For H2O GroupBy Q6 on medium dataset, initialization time is decreased from 170-200ms to 2-3ms (I guess mostly because we now don't use the default allocator from many threads simultaneously).


kurapov-peter (approved):

ienkovich:
I guess was wrong about the speed-up reason. We create a single RowSetMemoryOwner for all kernels, therefore lock on each individual Quantile object allocation becomes very costly.

It also means that we might have similar inefficiency in chunk allocations in quantile runtime. We use bigger chunk sizes there though but I'll try to check if we can improve multi-threaded memory allocations there.

PR #654, opened 2023-08-25
Avoid unused values materialization in ResultSet::rowCount.

ienkovich:
This is a minor optimization of a single-threaded rowCount implementation. It's not used for big result sets but still, we materialize TargetValue objects here and don't use them. Materialization can be costly for complex aggregates (like approx quanile and topK) and big varlen values.

### DIFF
--- a/omniscidb/ResultSet/ResultSet.cpp
+++ b/omniscidb/ResultSet/ResultSet.cpp
@@ -629,17 +629,12 @@ size_t ResultSet::rowCountImpl(const bool force_parallel) const {
   if (force_parallel || entryCount() >= auto_parallel_row_count_threshold) {
     return parallelRowCount();
   }
-  std::lock_guard<std::mutex> lock(row_iteration_mutex_);
-  moveToBegin();
   size_t row_count{0};
-  while (true) {
-    auto crt_row = getNextRowUnlocked(false, false);
-    if (crt_row.empty()) {
-      break;
-    }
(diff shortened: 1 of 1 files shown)


omniscidb/ResultSet/ResultSet.cpp
```diff
@@ -629,17 +629,12 @@ size_t ResultSet::rowCountImpl(const bool force_parallel) const {
...
-  std::lock_guard<std::mutex> lock(row_iteration_mutex_);
-  moveToBegin();
   size_t row_count{0};
-  while (true) {
-    auto crt_row = getNextRowUnlocked(false, false);
-    if (crt_row.empty()) {
-      break;
-    }
+  auto iter = rowIterator(false, false);
+  while (iter.isValid()) {
```
kurapov-peter, line 634: Should `isValid` be protected by a lock?

kurapov-peter (approved):

omniscidb/ResultSet/ResultSet.cpp
ienkovich: We have an internal iterator object and work with it should be protected by a lock. But here I create an external iterator to avoid locks.

BTW I really don't like this internal iterator. I barely can imagine cases when we cannot use an external iterator. And we regularly had problems with this internal one. E. g. debug print of the result set moved its internal iterator to the end and then we incorrectly iterated through it in some other place. We fixed multiple cases like that previously.

PR #656, opened 2023-08-29
Introduce per-thread memory pools for lock-free aloocations in quantile.

ienkovich:
In this patch, I'm trying to improve memory allocations for quantile aggregate. Currently, memory is allocated through the `RowSetMemoryOwner::allocate` method that locks a mutex for each call.

In the proposed code I use `tbb::this_task_arena::current_thread_index` to identify the thread, then allocate 10MB chunk of memory for it and subsequent allocations go with no synchronization at all until we run out of memory for this thread.

Perf results are really flaky for this test. I use 1005 server and run H2O GroupBy Q6 query on the 5GB dataset, making 9 query runs in the same process. The main branch shows 0.65-1.0s result, and the median is 0.75. With the patch, I have 0.55-0.9s, and the median is 0.63. The profit is not as big as I hoped it would be, but this query actually computes multiple aggregates and has two execution steps, so the kernel's speed-up is more significant.

No synchronization at all on memory allocation is achieved by a guarantee from a caller that the same thead_idx isn't used by different threads. This is achieved by always running kernels through TBB on a single TBB task arena, but the code would break if we started using quantile out of TBB context or with multiple task arenas.

I can add mutexes to per-thread memory pools to make it much safer. Misuse of the per-thread allocation would then just cause slow cod rather than memory corruption. But this would reduce the gain of the patch roughly by half for the Q6.

### DIFF
--- a/omniscidb/Shared/quantile.h
+++ b/omniscidb/Shared/quantile.h
@@ -11,6 +11,8 @@
 
 #include "IR/OpTypeEnums.h"
 
+#include <tbb/task_arena.h>
+
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
@@ -30,6 +32,8 @@ class ChunkedArray {
     size_t max_elems;
   };
 
+  using ChunkVector = std::vector<Chunk>;
+
   // Random access iterator to be used with std::nth_element.
   template <typename T>
   class Iterator {
@@ -40,7 +44,7 @@ class ChunkedArray {
     typedef T& reference;
     typedef std::random_access_iterator_tag iterator_category;
 
-    Iterator(const std::vector<Chunk>* chunks, size_t chunk_idx, size_t chunk_offs)
+    Iterator(const ChunkVector* chunks, size_t chunk_idx, size_t chunk_offs)
         : chunks_(chunks), chunk_idx_(chunk_idx), chunk_offs_(chunk_offs) {}
 
     Iterator(const Iterator& other) = default;
@@ -166,7 +170,7 @@ class ChunkedArray {
     }
 
    private:
-    const std::vector<Chunk>* chunks_;
+    const ChunkVector* chunks_;
     // Current chunk index. Can be equal to size of chunks_ vector for `end` iterator.
     size_t chunk_idx_;
     // Offset in the current chunk. Should always be less than chunk size when the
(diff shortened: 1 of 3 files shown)


alexbaden:
I'm ok with the approach. Is there any way to improve safety (i.e. somehow detect if there were multiple task arenas?). But, if it's robust if not misused then I think it's acceptable. 

We also have:
```
extern "C" RUNTIME_EXPORT void register_buffer_with_executor_rsm(int64_t exec,
                                                                 int8_t* buffer) {
  Executor* exec_ptr = reinterpret_cast<Executor*>(exec);
  if (buffer != nullptr) {
    exec_ptr->getRowSetMemoryOwner()->addVarlenBuffer(buffer);
  }
}
```
which is used for some varlen operations. It is doing roughly the same thing, though the memory is allocated by the kernel dynamically, then just attached to the RowSetMemoryOwner to keep track of it. Does decoupling allocation + registration of the buffer with RSM help you at all?

ienkovich:
> I'm ok with the approach. Is there any way to improve safety (i.e. somehow detect if there were multiple task arenas?)

I don't think we can identify from the quantile code that different kernels are not executed in different task arenas. The code has to be robust if a single arena is used. It doesn't have to be a global arena, just the same arena for all kernels using the same allocator. TBB guarantees the uniqueness of provided thread indices. If we run outside of any task arena then -1 is returned which should cause regular `allocate` to be used.

We will probably want to use multiple arenas someday for better NUMA management. TBB now supports task arenas pinned to the specified NUMA node. But in this case, it's probably reasonable to also use multiple allocators to have a separate `folly::Arena` for each NUMA node.

> Does decoupling allocation + registration of the buffer with RSM help you at all?

I doubt `malloc` can be efficient enough in a multi-threaded environment (we can use TBB version of malloc though). Another reason I don't want to decouple allocation from the RowSetMemoryOwner is a memory arena. We make millions of allocations throughout kernels execution. Calling free on each allocated chunk afterward can be costly, with memory arena deallocation is free. But that's just a theory, I didn't run experiments with malloc+free option. BTW registration shouldn't be required in this case since we call destructors for all quantiles and we could free the memory there. Also, registration from multiple threads would require some locks I'm trying to avoid.

I can try to use `scalable_` allocators from TBB instead of an Arena to see if it works better.

ienkovich:
I checked it and using TBB's `scalable_malloc` instead of `RowSetMemoryOwner` in quantile actually leads to ~2x performance degradation (compared to the main branch). So we better keep using folly Arena.

ienkovich:
@alexbaden @kurapov-peter Is it OK to merge this one to finish with the quantile series?

omniscidb/Shared/quantile.h
```diff
@@ -180,11 +184,16 @@ class ChunkedArray {
   void push(T value) {
     // Check if we need to allocate a new chunk.
     if (chunks_.empty() || cur_idx_ == chunks_.back().max_elems) {
-      // Allocator is most probably a RowSetMemoryOwner object. It is not supposed to be
-      // used to allocate very small objects, so we start with 1 KB and double it each
-      // time with 64KB limit.
+      if (thread_idx_ < 0) {
+        thread_idx_ = tbb::this_task_arena::current_thread_index();
+      } else if (thread_idx_ != tbb::this_task_arena::current_thread_index()) {
```
kurapov-peter, line 189: Looks like you might be able to not store `thread_idx_` at all since `current_thread_index` is queried always anyway (unless I didn't notice it somewhere else?).

kurapov-peter (approved):

omniscidb/Shared/quantile.h
ienkovich: It is stored to make sure that `push` method of the same object is not called from different threads. It might help to detect threading issues that would be hard to debug otherwise and I also didn't see any significant performance impact of this check so let it be.

PR #657, opened 2023-08-31
Added QUANTILE agg function to Calcite

AndreyPavlenko:


ienkovich (commented):
Is this declaration good enough to handle interpolation methods supported by `QUANTILE`? Also, `QUANTILE` supports datetime data types.

AndreyPavlenko:
Added the interpolation parameter, datetime data is also supported, the test passes - https://github.com/modin-project/modin/pull/6528/files#diff-924b045aa089c05b3a717bc9cea19dc8581e2b8351fcc3c0568a54a27ce8a146R1284

PR #660, opened 2023-09-06
Update to LLVM 15 and Support LLVM 16 on Windows

alexbaden:
Updates CI to use LLVM 15 and adds build support for LLVM 16.

For CI, the main change is to set VSINSTALLDIR so LLVM picks it up for proper linking. To do this, I use the `vswhere.exe` application which is installed with visual studio. Presumably this will survive version updates. I also modified the cache getters/setters as the cache was not being used in the build step, previously. Finally, my initial plan was to go to LLVM 16 but I ran into some opaque pointer issues even with my workarounds, so we're at LLVM 15 but I kept the commits to add build support for 16. 

Interestingly my first run through the tests failed because I hit the same relocatable section error we've been seeing, so unfortunately LLVM 15 isn't fixing that one. Maybe 16 will.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -11,7 +11,6 @@ on:
       cuda_compiler_version:
         type: string
 
-
     outputs:
       cache-suffix:
         value: ${{ jobs.build.outputs.cache-suffix }}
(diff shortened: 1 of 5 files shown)


gshimansky (approved):

kurapov-peter (approved):

leshikus (commented):
LGTM

Issue #588, opened 2023-07-19
Execute.cpp:4306 Check failed: dd && dd->stringDict 


AndreyPavlenko:
The following Modin code:
```python
import modin.pandas as pd

data = range(100)
df = pd.DataFrame({"A": data})
df["B"] = "data"
df["C"] = pd.Series(data)
repr(df)
```
fails with error:
```
Execute.cpp:4306 Check failed: dd && dd->stringDict 
```

Probably, the following failure is also related to this one:
```python
import modin.pandas as pd

data = range(100)
df = pd.DataFrame({"A": data})
for i in range(3):
    df[f"B{i}"] = pd.Series(data)
    repr(df)
```
```
TableGenerations.cpp:23 Check failed: it_ok.second
```

alexbaden:
@AndreyPavlenko did this just start happening recently (i.e. did this query used to work)?

AndreyPavlenko:
> @AndreyPavlenko did this just start happening recently (i.e. did this query used to work)?

I don't know, I didn't check it before.

PR #661, opened 2023-09-07
Replace dictionary proxies with nested dictionaries 01/N

ienkovich:
This patch series aims to remove the `StringDictionaryProxy` class by extending `StringDictionary` functionality to cover all use cases of the proxy class. The reason is that the dictionary encoding used for ResultSets is incompatible with the dictionary encoding used by ArrowStorage and expected by execution kernels. It means we simply cannot use dictionary-encoded columns from ResultSets and need to convert them (currently, ResultSetRegistry lacks such functionality, see #588). Instead of implementing such a conversion, I want to change the coding so that we wouldn't need to convert anything.

There are a few major changes required to achieve this:
1. Stop using negative indices in dictionary-encoded columns in ResultSets. It is possible because proxies already use dictionary generations, i.e. max index value that can be used to address the base dictionary strings. Therefore we can simply use higher values to encode strings owned by the proxy.
2. Enable deep nesting levels. Since we can use execution results as a query input, we need the ability to use an existing dictionary hierarchy as a base for a new dictionary. This means either proxy and dictionary should have the same API, or we simply use a single class (I chose to use StringDictionary only)
3. Switch to a single API dictionary in storage and execution to smoothly run queries on execution results

I try to make changes in many small steps to simplify the review process. So, many patches are very trivial refactoring or a small part of the new functionality.

This patch removes unused StringDictionaryProxy code.


alexbaden (approved):

omniscidb/StringDictionary/StringDictionaryProxy.cpp
```diff
@@ -534,19 +519,6 @@ class StringLocalCallback : public StringDictionary::StringCallback {
   }
 };
 
-// Union strings from both StringDictionaryProxies into *this as transients.
-// Return id_map: sdp_rhs:string_id -> this:string_id for each string in sdp_rhs.
-StringDictionaryProxy::IdMap StringDictionaryProxy::transientUnion(
```
alexbaden, line 539: Surprised that this is unused.

Issue #648, opened 2023-08-24
device-linear-multifrag execution mode


akroviakov:
As of now, HDK's heterogeneity looks like this:

- We have X fragments, when we schedule them on a GPU, it will receive X kernels, X fragments and execute kernels sequentially.

The currently disabled multifragment option does the following:

- We have X fragments, when we schedule them on a GPU, it will receive 1 kernel and X fragments and execute this one kernel.

Why is current multifragment good: 
- We avoid repeated administrative work around preparing X kernels, scheduling them and retrieving results. 
- Fine grained control (allocate/free) over memory of fragments.

Why is it not _that_ good:

- Currently we place fragments in GPU memory arbitrarily, this is taken care of in the code of GPU kernels which increases their complexity for the programmer, we cannot simply index into the multifragmented column and we have to do more checks per tuple (e.g.,  `JoinColumnIterator& operator++()` and related for loops). This likely eats some performance from the `executePlan` stage, which is often the costliest.
_____________

**Basic idea:**
The best case for the kernel is if all columns are linear (simple loops over dense data). All GPU-assigned fragments have to be _materialized anyways, so why not place them linearly_? 
We can add a new execution mode where fragments (regardless of their position in host memory) are assembled linearly on GPU, this way, the fragments become transparent for the GPU kernel, it will treat them as one big fragment. We still know the address of each fragment and can get it back to CPU, but we cannot free individual fragments, only if we do it for the whole linear buffer. Such behavior is not different from just a big fragment size. 

**Why is it supposed to be better?**
- Cheap initial data locality
- Easier indexing into a column -> easier and (likely) faster kernel code

**What's the catch?**
- We lose fine granularity of memory control on GPU. But do we _always care_?
- What if the next step takes more fragments? We know which fragments are still on GPU and their location, so one could avoid using bandwidth by memmove/memcpy'ing on the device itself and add only missing ones. If it is not possible, we pay the fetching price. Maybe this is a subject to more sophisticated data structures.

This won't bring any benefit for CPU as it is expensive and redundant (with regards to memory) to linearize fragments, also likely pointless anyways, since each thread processes one fragment (which is linear) at a time.

**When does this execution mode make sense:**
- Query is just one step (or more with the latter ones being more compute intense)
- Proportions do not change between steps (this is in a simple CPU+GPU scenario, but the more devices we have, the more likely we are to shuffle the fragments between them anyways)
- Pure plan execution speed, the workload is not severely bandwidth bound. 

[...]

PR #665, opened 2023-09-14
Get rid of frag count overhead in prepareKernelParams

akroviakov:
Imagine you read CSVs into a table (e.g., 80Mil. rows) and you want to "try" to reduce materialization overhead, so you pick a small fragment size (e.g., 40k) or you somehow managed to operate on arrow chunks directly (roughly 50k rows, zero copy for CPU). You know that you are in a heterogeneous setting, so devices can actually share a workload since there are many fragments and you can also do multifragment on GPU. You do not expect any issues. That is until you try to run 8 subsequent group by count on GPU each on a different column. You will likely find out that **fetching a column** and **preparing kernel parameters** gradually takes more and more time and at 8th run each of them takes **over 400ms (30% of the whole query)**. A new bottleneck emerges due to a higher fragment count,  which leads to more metadata being passed (i.e., more allocations and data transfers). This might have been OK if it wasn't for  **preparing kernel parameters** being **per kernel (i.e., step) overhead**. This wasn't an issue because no one would do 40k fragment size on GPU :), but in a heterogeneous setup it might be the case.

This PR does one allocation for all of the kernel parameters and fills the memory by asynchronous (for CUDA) data transfer calls. Why async? Well, why not, we have potentially many independent data transfers that need to be ready only before kernel launch. This PR reduces **preparing kernel parameters** overhead to basically only data transfers (which in practice almost always becomes **0ms**). 

That is, in the example above (8 subsequent group bys) the gradual increase of **preparing kernel parameters** overhead is eliminated.

- The main benefit of course comes from one allocation instead of linear-to-frag-count allocations.
- Maybe this small introduction to async data transfer capability might also be useful elsewhere.
- Also since kernel parameters do not live past `launchGpuCode()` and now they are a single buffer, we can now easily free it, instead of piling up all those allocations.

### DIFF
--- a/omniscidb/QueryEngine/QueryExecutionContext.h
+++ b/omniscidb/QueryEngine/QueryExecutionContext.h
@@ -125,7 +125,16 @@ class QueryExecutionContext : boost::noncopyable {
     KERN_PARAM_COUNT,
   };
 
-  std::vector<int8_t*> prepareKernelParams(
+  size_t getKernelParamsAllocSize(
+      const std::vector<std::vector<const int8_t*>>& col_buffers,
+      const std::vector<int32_t>& error_codes,
+      const size_t literals_sz,
+      const std::vector<std::vector<int64_t>>& num_rows,
+      const std::vector<std::vector<uint64_t>>& frag_offsets,
+      const size_t init_agg_vals_sz,
+      const size_t hash_table_count) const;
+
+  std::pair<std::vector<int8_t*>, Data_Namespace::AbstractBuffer*> prepareKernelParams(
       const std::vector<std::vector<const int8_t*>>& col_buffers,
       const std::vector<int8_t>& literal_buff,
       const std::vector<std::vector<int64_t>>& num_rows,
--- a/omniscidb/BufferProvider/BufferProvider.h
+++ b/omniscidb/BufferProvider/BufferProvider.h
@@ -34,6 +34,17 @@ class BufferProvider {
                             const int8_t* host_ptr,
                             const size_t num_bytes,
                             const int device_id) const = 0;
+
+  virtual void copyToDeviceAsyncIfPossible(int8_t* device_ptr,
+                                           const int8_t* host_ptr,
+                                           const size_t num_bytes,
+                                           const int device_id) const = 0;
+
+  virtual void copyToDeviceAsync(int8_t* device_ptr,
+                                 const int8_t* host_ptr,
+                                 const size_t num_bytes,
+                                 const int device_id) const = 0;
+  virtual void synchronizeStream(const int device_id) const = 0;
   virtual void copyFromDevice(int8_t* host_ptr,
                               const int8_t* device_ptr,
                               const size_t num_bytes,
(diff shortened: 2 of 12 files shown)


ienkovich:
Why do we have many kernels on GPU? If we use multi-frag kernels, then increasing the number of fragments x1000 shouldn't increase the number of kernels, it should increase the number of fragments per kernel instead. What do I miss?

akroviakov:
You are correct. One multifrag kernel has to have info on all fragment it touches. In kernel-per-fragment it is only one fragment and the overhead is not noticeable. Assume you have have one kernel with 1000 fragments, if you take a look at the current loop in `prepareKenelParams()`, it would do 1000 GPU allocations for `#cols pointers`, this kills the performance. On top of that, we pass `#cols pointers` for each of 1000 fragments synchronously. 

For example, currently in group by count with one column and 1000 fragments, we get **1000 separate GPU allocations of size 8 bytes** and after each allocation we synchronously copy the pointer itself, so 1000 sync 8 bytes data transfers to GPU. After a kernel is done there is no cleanup of what `prepareKenelParams()` has allocated (I haven't seen any), so they pile up and GPU seems to not like it.

I wrote

>  per kernel (i.e., step) overhead.

to indicate that this overhead is not like fetching a data (data remains on GPU for next queries) and is not even like running a query, because in a multi-step query you would experience this overhead `#steps` times.

In fact, there are quite a few other places with per query overhead due to where many fragments, currently known ones are:
- create QueryExecutionContext (for GPU) : [we are creating result sets in a single thread loop which copies fragment info](https://github.com/intel-ai/hdk/blob/585911be950fbbd58faee29f0c9981d7f18d9960/omniscidb/QueryEngine/QueryMemoryInitializer.cpp#L427)
- Query pre-execution steps (for CPU and GPU): [for each table we copy fragments and chunk keys](https://github.com/intel-ai/hdk/blob/585911be950fbbd58faee29f0c9981d7f18d9960/omniscidb/QueryEngine/Execute.cpp#L4282)

ienkovich:
Thanks for clarifications!

Does GPU benefit of having many small fragments? For CPU it might allow to zero-copy fetch data. But for GPU we have to copy it anyway, so we might always to make it a single linear fragment while copying.

omniscidb/QueryEngine/QueryExecutionContext.h
```diff
@@ -125,7 +125,16 @@ class QueryExecutionContext : boost::noncopyable {
     KERN_PARAM_COUNT,
   };
 
-  std::vector<int8_t*> prepareKernelParams(
+  size_t getAllocSizeKernelParams(
```
ienkovich, line 128: `getKernelParamsSize` or `getKernelParamsAllocSize` sound better to me.

ienkovich (approved):

akroviakov:
> Does GPU benefit of having many small fragments?

GPU kernel execution on itself does not benefit from many fragments, but it also doesn't suffer much. Some queries can be accelerated by heterogeneous execution with a small enough fragment size. To increase the number of "heterogeneous cases", we have to make HDK (e.g., its GPU code path) more friendly to small fragments (i.e., more fragments).

> make it a single linear fragment while copying

Some thoughts and issues of this were expressed here: #648.

kurapov-peter (approved):

PR #669, opened 2023-09-20
Reduce copy for result sets in QueryMemoryInitializer

akroviakov:
This PR is another step in making HDK more fragment transparent. GPU in multifrag mode experienced a fragment-count dependent overhead in `QueryMemoryInitializer()`.
Example of a bad scenario: single threaded loop does 130 iterations, each iteration should copy `col_buffers`, `column_frag_offsets`, `column_frag_sizes`. With huge fragments the overhead is barely observable, however lowering the fragment size (e.g., to accommodate for CPU parallelism, split workload) makes it noticeable.
There are cases where the time we spend in `create QueryExecutionContext` is **20% of the entire query** (e.g., 80 Mil.rows with only **16 fragments**, perform group by count on GPU).

This PR eliminates the described fragment-count dependent overhead. Example: `create QueryExecutionContext` takes **1ms for 3000 fragments**.

Pitfalls?
The fragment metadata pieces targeted by this PR seem to be immutable in `ResultSet`, they are only used for appending another `ResultSet` (combine pieces of metadata without modifying either) and iteration, so why not "share" these metadata pieces across result sets?

This "sharing" begins (we actually make a safe copy of the required metadata) right before the loop. This means that `ResultSet` does not reference the current state of a table, but we still avoid copying for 2nd, 3rd, ..., Nth `ResultSet`.


omniscidb/QueryEngine/QueryExecutionContext.cpp
```diff
@@ -124,9 +124,9 @@ ResultSetPtr QueryExecutionContext::groupBufferToDeinterleavedResults(
   auto deinterleaved_result_set =
       std::make_shared<ResultSet>(result_set->getTargetInfos(),
                                   std::vector<ColumnLazyFetchInfo>{},
-                                  std::vector<std::vector<const int8_t*>>{},
-                                  std::vector<std::vector<int64_t>>{},
-                                  std::vector<int64_t>{},
+                                  std::make_shared<const ColBuffers>(),
```
ienkovich, line 127: Why not simply `nullptr`? Column buffers are supposed to be used for lazy columns only.

omniscidb/ResultSet/ResultSet.h
```diff
@@ -176,9 +183,9 @@ class ResultSet {
 
   ResultSet(const std::vector<TargetInfo>& targets,
             const std::vector<ColumnLazyFetchInfo>& lazy_fetch_info,
-            const std::vector<std::vector<const int8_t*>>& col_buffers,
-            const std::vector<std::vector<int64_t>>& frag_offsets,
-            const std::vector<int64_t>& consistent_frag_sizes,
+            const ColBuffersPtr col_buffers,
```
ienkovich, line 186: const doesn't match the definition

omniscidb/ResultSet/ResultSet.cpp
```diff
@@ -84,9 +84,9 @@ ResultSet::ResultSet(const std::vector<TargetInfo>& targets,
 
 ResultSet::ResultSet(const std::vector<TargetInfo>& targets,
                      const std::vector<ColumnLazyFetchInfo>& lazy_fetch_info,
-                     const std::vector<std::vector<const int8_t*>>& col_buffers,
-                     const std::vector<std::vector<int64_t>>& frag_offsets,
-                     const std::vector<int64_t>& consistent_frag_sizes,
+                     ColBuffersPtr col_buffers,
```
ienkovich, line 87: These now can be moved to appropriate fields rather than copied.

omniscidb/QueryEngine/QueryMemoryInitializer.cpp
```diff
@@ -383,6 +383,12 @@ QueryMemoryInitializer::QueryMemoryInitializer(
     group_by_buffers_.push_back(varlen_output_buffer);
   }
 
+  ColBuffersPtr res_set_col_buf = std::make_shared<const ColBuffers>(col_buffers);
+  FragOffsetsPtr res_set_col_frag_offsets = std::make_shared<const FragOffsets>(
+      get_col_frag_offsets(ra_exe_unit.target_exprs, frag_offsets));
+  FragSizesPtr res_set_frag_sizes = std::make_shared<const FragSizes>(
+      get_consistent_frags_sizes(ra_exe_unit.target_exprs, consistent_frag_sizes));
```
ienkovich, line 390: BTW I suspect that these calls moved out of the loop can be more important for perf than vector copies.

ienkovich (approved):
Looks good! There are few minors though.

omniscidb/QueryEngine/QueryMemoryInitializer.cpp
akroviakov: It didn't help much, `shared_ptr` was what brought things to ~1ms

omniscidb/QueryEngine/QueryExecutionContext.cpp
```diff
@@ -124,9 +124,9 @@ ResultSetPtr QueryExecutionContext::groupBufferToDeinterleavedResults(
   auto deinterleaved_result_set =
       std::make_shared<ResultSet>(result_set->getTargetInfos(),
                                   std::vector<ColumnLazyFetchInfo>{},
-                                  std::vector<std::vector<const int8_t*>>{},
-                                  std::vector<std::vector<int64_t>>{},
-                                  std::vector<int64_t>{},
+                                  nullptr,
+                                  std::make_shared<const FragOffsets>(),
```
ienkovich, line 128: All of them can be nullptr.

PR #671, opened 2023-09-21
Reduce query pre-execution steps copies

akroviakov:
This PR is another step in making HDK more fragment transparent.

Target case: table with many columns (over 20) split into considerable number of fragments.

Problem: the entire table info (size depends on fragment count) is copied regardless of query-relevant columns.

Solution: since `InputTableInfoCache` is scoped to `executeRelAlgQueryNoRetry()` and the only place where anything is inserted into the cache (and never updated) is `InputTableInfoCache::getTableInfo()` , can't we just const reference it (especially for `computeTableGenerations()` where we only need tuple count)?

Effect: well, we still have to materialize `InputTableInfo` in cache, but we can save on further copies, so **3x faster `Query pre-execution steps`**.

### DIFF
--- a/omniscidb/ArrowStorage/ArrowStorage.h
+++ b/omniscidb/ArrowStorage/ArrowStorage.h
@@ -170,6 +170,8 @@ class ArrowStorage : public SimpleSchemaProvider, public AbstractDataProvider {
     std::vector<DataFragment> fragments;
     size_t row_count = 0;
     TableStats table_stats;
+    std::shared_ptr<const TableFragmentsInfo> metadata =
+        std::make_shared<TableFragmentsInfo>();
   };
 
   struct DictionaryData {
@@ -243,6 +245,7 @@ class ArrowStorage : public SimpleSchemaProvider, public AbstractDataProvider {
                             size_t num_bytes) const;
   void refragmentTable(TableData& table, const int table_id, const size_t new_frag_size);
   void materializeDictionary(DictionaryData* dict_data);
+  void setTableMetadata(TableData& table, const int table_id) const;
 
   int db_id_;
   int schema_id_;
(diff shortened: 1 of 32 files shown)


akroviakov:
> We can even change AbstractBufferMgr::getTableMetadata to return `shared_ptr`, which would avoid the creation of a new structure on each call to the storage

Do you mean introducing a storage-level cache that stores `shared_ptr` to table's metadata and on subsequent calls just returns that `shared_ptr`?

ienkovich:
> Do you mean introducing a storage-level cache that stores `shared_ptr` to table's metadata and on subsequent calls just returns that `shared_ptr`?

Yes. That doesn't have to be a part of this PR though. It can be a further optimization if you feel it can give a significant gain.

omniscidb/ArrowStorage/ArrowStorage.h
```diff
@@ -73,7 +73,8 @@ class ArrowStorage : public SimpleSchemaProvider, public AbstractDataProvider {
   std::unique_ptr<Data_Namespace::AbstractDataToken> getZeroCopyColumnData(
       const ColumnRef& col_ref) override;
 
-  TableFragmentsInfo getTableMetadata(int db_id, int table_id) const override;
+  const std::shared_ptr<TableFragmentsInfo> getTableMetadata(int db_id,
```
ienkovich, line 76: What is this `const` for here and all other places? Did you mean to have `std::shared_ptr<const TableFragmentsInfo>`?

omniscidb/DataMgr/BufferMgr/BufferMgr.h
```diff
@@ -180,9 +180,10 @@ class BufferMgr : public AbstractBufferMgr {  // implements
     return nullptr;
   }
 
-  TableFragmentsInfo getTableMetadata(int db_id, int table_id) const override {
+  std::shared_ptr<const TableFragmentsInfo> getTableMetadata(int db_id, int table_id)
+      const override {
     UNREACHABLE();
-    return TableFragmentsInfo{};
+    return std::make_shared<TableFragmentsInfo>(TableFragmentsInfo{});
```
kurapov-peter, line 186: Looks like an unnecessary copy constructor call. Did you mean to write `std::make_shared<TableFragmentsInfo>()`?

kurapov-peter (approved):
Looks reasonable.

kurapov-peter (approved):

PR #672, opened 2023-09-22
Backport release changes

leshikus:
Update recipe, add rev-parse fix, zstd guard fix


recipe/meta.yaml
```diff
@@ -1,6 +1,6 @@
-{% set version = "0.6.0" %}  # PEP 386
+{% set version = "0.8.0" %}  # PEP 386
```
Garra1980, line 1: it should be 0.9.0 I guess

leshikus: thanks; this does not actually affect the nightly build, it uses main branch and not a source release

leshikus: fixed

alexbaden (approved):
Looks good!

PR #675, opened 2023-09-25
Replace dictionary proxies with nested dictionaries 04/N

ienkovich:
Rename StringDictionaryProxy get/add transient string methods to match the dictionary interface.


alexbaden (approved):
I suppose it's clear from the context when we have a transient string, but the previous interface was nice because we knew these strings were temporary.

ienkovich:
> but the previous interface was nice because we knew these strings were temporary.

This is an actual problem. We treat them as some special kind of strings, but we need them to be regular ones when we try to run queries on ResultSets.

PR #678, opened 2023-09-27
Replace dictionary proxies with nested dictionaries 06/N

ienkovich:
Add `StringDictionaryProxy::copyStrings`. This is to avoid special handling of transient strings (special ID handling will be removed later).

### DIFF
--- a/omniscidb/StringDictionary/StringDictionaryProxy.h
+++ b/omniscidb/StringDictionary/StringDictionaryProxy.h
@@ -180,10 +180,12 @@ class StringDictionaryProxy {
   // in the std::map when new strings are added, but will change in a std::vector.
   using TransientMap = std::map<std::string, int32_t, std::less<>>;
 
-  const std::vector<std::string const*>& getTransientVector() const {
-    return transient_string_vec_;
-  }
+  std::vector<std::string> copyStrings() const;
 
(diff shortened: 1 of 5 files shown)


omniscidb/StringDictionary/StringDictionaryProxy.h
```diff
@@ -180,10 +180,12 @@ class StringDictionaryProxy {
   // in the std::map when new strings are added, but will change in a std::vector.
   using TransientMap = std::map<std::string, int32_t, std::less<>>;
 
-  const std::vector<std::string const*>& getTransientVector() const {
-    return transient_string_vec_;
-  }
+  std::vector<std::string> copyStrings() const;
 
+  // Iterate over transient strings, then non-transients.
+  void eachStringSerially(StringDictionary::StringCallback&) const;
```
kurapov-peter, line 186: Is the iteration order significant?

kurapov-peter (approved):

omniscidb/StringDictionary/StringDictionaryProxy.h
ienkovich: I don't think we provide any particular order. And this API is not actually used anyway. Method is moved to match the Dictionary API.

PR #679, opened 2023-09-27
[Join] Buffer free

Devjiu:
This commit refactors and simplifies method `getAllTableColumnFragments`. Also some parallelization added.

Partially resolves: #574


omniscidb/QueryEngine/ColumnFetcher.cpp
```diff
@@ -478,6 +481,18 @@ const int8_t* ColumnFetcher::linearizeColumnFragments(
   // prepare ChunkIter for the linearized chunk
   auto merged_chunk = std::make_shared<Chunk_NS::Chunk>(
       merged_data_buffer, merged_index_buffer, col_info);
+  merged_data_buffer->deleteWhenUnpinned();
```
ienkovich, line 484: It's enough to call this method once (in dtor).

Devjiu: done, review again, please

Devjiu:
@alexbaden @kurapov-peter PTAL, this pr changes free method contract a little, and also changes interaction of buffer with bufferManager. It works, but perhaps you have some recommendations about method's visibility and class responsibilities.

PR #680, opened 2023-09-28
Replace dictionary proxies with nested dictionaries 07/N

ienkovich:
Avoid negative generation in dictionary proxy. There is no strong reason to create a proxy with a negative generation anyway. Non-zero generation on a proxy creation would allow us later to stop using negative indices for transient strings. Non-negative indices would allow deep nesting and let us treat proxy-encoded columns as regular dictionary-encoded columns.

### DIFF
--- a/omniscidb/StringDictionary/StringDictionaryProxy.cpp
+++ b/omniscidb/StringDictionary/StringDictionaryProxy.cpp
@@ -42,7 +42,13 @@ constexpr int32_t transient_id_ceil{-2};
 
 StringDictionaryProxy::StringDictionaryProxy(std::shared_ptr<StringDictionary> sd,
                                              const int64_t generation)
-    : string_dict_(sd), generation_(generation) {}
+    : string_dict_(sd), generation_(generation) {
+  if (generation_ < 0) {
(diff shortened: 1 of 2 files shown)


omniscidb/StringDictionary/StringDictionaryProxy.cpp
```diff
@@ -42,7 +42,13 @@ constexpr int32_t transient_id_ceil{-2};
 
 StringDictionaryProxy::StringDictionaryProxy(std::shared_ptr<StringDictionary> sd,
                                              const int64_t generation)
-    : string_dict_(sd), generation_(generation) {}
+    : string_dict_(sd), generation_(generation) {
+  if (generation_ < 0) {
+    generation_ = static_cast<int64_t>(sd->storageEntryCount());
```
kurapov-peter, line 47: Do you plan to forbid negative values later on?

kurapov-peter (approved):

omniscidb/StringDictionary/StringDictionaryProxy.cpp
ienkovich: Yes, it will be in the next patch.

PR #686, opened 2023-10-04
Refactor tests

leshikus:
Subj


.github/workflows/main.yml
```diff
@@ -175,17 +184,9 @@ jobs:
...
     with:
       cache-suffix: ${{ needs.build-windows.outputs.cache-suffix }}
       test: sanity
-
-  modin:
-    name: Integration HDK with Modin (Pytest)
-    uses: ./.github/workflows/modin.yml
-
-  pytest:
-    name: PyHDK (Pytest)
```
alexbaden, line 190: I don't see where the HDK Python tests are being run, now?

alexbaden (changes requested):
Two comments:

1) We need more description about why these changes are happening. To me, they are not self explanatory. Reducing churn in the job names is important so developers can quickly understand what is going wrong if a test fails, so if we change job names we need to justify it.

2) See inline comment on removing running pyhdk pytest. I am open, however, to potentially merging the existing `pytest` run and the modin run similar to how you have modified the `test.yml` file. There is probably an opportunity to reduce extra builds here.

.github/workflows/main.yml
```diff
@@ -65,16 +65,25 @@ jobs:
...
       test: sanity
 
+  test-modin:
+    name: Python tests
+    needs: build
+    uses: ./.github/workflows/test.yml
+    with:
+      name: 'gcc-cpu'
+      cache-suffix: ${{ needs.build.outputs.cache-suffix }}
+      test: modin
```
Devjiu, line 83: With single job we actually mixing pyhdk api test and modin check. @ienkovich  PTAL. 
I think it's fine, but there may be a need to see them separately.

.github/workflows/main.yml
Devjiu: It's in `.github/workflows/pytest.yml:64` below (next level job). All jobs in main.yml if they don't `needs` any triggered simultaneously on pull request.

leshikus:
@alexbaden there are multiple matters

1) the recent version of `conda run` has a bug that it can be broken by some specific option even if this option is a part of a command we want to run; thus I moved commands into dedicated scripts; this also adds to readability

2) I wanted to remove code and build duplication for some time, so I finally made this along the previous change

3) I combined pytest and modin tests into one task; this is just a matter of taste; we may discuss further the name for a combined task or I can revert the change

4) I removed duplicated runs for `python python/tests/test_pyhdk_bindings.py` and `python python/tests/test_pyhdk_sql.py`; actually it was I who added them on Igor's request before `pytest.yml` was created so I have a responsibility to remove them if they are no longer needed; if they are needed I'll add them back, they run second time in a slightly different environment with modin installed; the tests from `pytest.yml` are moved as is into `test-modin.sh`

5) I replaced "Gtests (Sanity)" with "Gtests"  everywhere - this is also subjective and my variant is shorter

alexbaden:
> 3. I combined pytest and modin tests into one task; this is just a matter of taste; we may discuss further the name for a combined task or I can revert the change; the tests from both `pytest.yml` and `modin.yml` were moved as is into `test-modin.sh` with the following exception (4)

I see... the file should probably be renamed, then. 

Regardless - to me, this is a refactoring of existing jobs that is not strictly better than what we had before, with no functional changes (except perhaps merging the HDK + Modin tests, which is still a little unclear - @Garra1980 thoughts?). I also would tend to disagree that moving parts of the test to bash scripts improves readability. To me, it means I need to now look in two places to see what a test is doing, instead of just one.

leshikus:
@alexbaden that's a good question about readability, let me share my thoughts

1) the idea is to combine multiple lines of code into functional unit which have a descriptive name; 
2) replacing two 200+ line files with one ~10 line file improves things as well

while readability question is subjective, another point has a stronger justification; we know now that conda run command line is buggy (`-vx`), thus we need to keep it minimal

leshikus:
@alexbaden let me elaborate that second reason a bit

if we will use `conda run -n omnisci-dev sh script_name.sh` in most places, we ensure their command line parser have less chances to parse our directives incorrectly

leshikus:
@Garra1980 told me, "do not change anything"; closing this PR

PR #687, opened 2023-10-04
Fix conda in a minimal way

leshikus:
Subj

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Configure and build
         run: |
           conda info --env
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -99,7 +99,7 @@ jobs:
       - name: Configure and build
         run: |
           conda info --env
-          conda run --no-capture-output -n omnisci-dev sh -vx omniscidb/scripts/conda/build.sh ${{ inputs.options }}
```
Devjiu, line 102: Do we need to remove `-vx` in most of scripts?

alexbaden (approved):
Seems to fix the build, so let's get it merged. Any idea what changed in conda to break this flag?

leshikus:
I removed `-vx` because our script calls `set -vx` itself

PR #690, opened 2023-10-06
Update to Calcite 1.35 (latest version)

alexbaden:
The primary changes to Calcite were in the parser. Some appear to be bug fixes (the changes around Timestamp casts make sense logically, whereas before we seemed to be inventing precision we didn't have). Others appear to be semantics (the cast removal in the interop query) that do not effect us, as the cast in that query appears to be a workaround. I believe the new aggregate call constructor is specified properly, but we will want to sanity test with modin after this is merged.

I updated some Java packages associated with Calcite but left most of those at their current versions pending @vlad-penkin's analysis.

### DIFF
--- a/omniscidb/Calcite/java/calcite/pom.xml
+++ b/omniscidb/Calcite/java/calcite/pom.xml
@@ -60,7 +60,7 @@
     <dependency>
       <groupId>org.apache.calcite</groupId>
       <artifactId>calcite-core</artifactId>
-      <version>1.33.0</version>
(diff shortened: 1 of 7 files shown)


alexbaden:
@ienkovich @vlad-penkin any comments?

ienkovich (approved):
Test changes look reasonable and I don't think Modin is going to be affected. It shouldn't rely on implicit conversions to be done by Calcite.

PR #694, opened 2023-10-10
Add ASAN runner

leshikus:
Subj


leshikus:
let me add some explanations about test-docker.yml fixes; the glados runner works in a way that I needed to pass proxy to the internal docker; we had several places were internal docker was built and launched and proxy should be passed, I've combined them

alexbaden (approved):

PR #714, opened 2023-11-17
Use HDK class in examples + some new examples

akroviakov:
This PR introduces the following example usages:
- Change `basic_python_usage.ipynb` to use `HDK` class (now used in all examples).
- Taxi example simplified due to `HDK` class usage.
- Manually construct query graphs. 
- - Showcase the benefit of compilation.
- - Show how one can run a query graph on CPU/GPU/some part on CPU and some part on GPU.
- Allow relatively easy table generation for heterogeneous aggregation performance analysis (with plots).

Some minor updates to `hdk.py`: allow table refragmenting and easy gpu memory cleanup.


kurapov-peter:
Looks good. Please check the `debug_logs` usage.

python/pyhdk/hdk.py
```diff
@@ -2094,8 +2094,8 @@ def device_type(self, value):
 
 class HDK:
     def __init__(self, **kwargs):
-        if "debug_logs" in kwargs:
```
ienkovich, line 2097: Why do you need this change? `log_severity` allows you to choose any logging level. `debug_logs` always enables max debug log level.

python/pyhdk/hdk.py
```diff
@@ -2500,6 +2500,20 @@ def sql(self, sql_query, query_opts=None, **kwargs):
         res.scan = self.scan(res.table_name)
         return res
 
+    def clear_gpu_mem(self):
+        """
+        Clears GPU memory of all previously transferred buffers.
+        """
+        self._executor.clearMemory(self._data_mgr, 2)
+
+    def create_refragmented_view(self, table_name, refragmented_view_name, frag_size):
```
ienkovich, line 2509: To be more consistent with the rest of API, I propose a few things:
1. Make view name optional here
2. Return a scan node of the created view that can be used for queries
3. Accept both table_name and scan node as input
4. Add similar interface to the scan node to move it there completely
```
ht1 = hdk.import_csv(...)
hdk.sql("SELECT ... FROM t;", t = ht1)
ht2 = ht1.create_refragmented_view(1000000)
hdk.sql("SELECT ... FROM t;", t = ht2)
hdk.drop_table(ht1)
hdk.drop_table(ht2)
```

python/pyhdk/hdk.py
```diff
@@ -2500,6 +2500,20 @@ def sql(self, sql_query, query_opts=None, **kwargs):
...
+        """
+        Creates a refragmented view of an existing table.
+        """
```
ienkovich, line 2512: Parameters and return value descriptions are missing.

python/pyhdk/hdk.py
akroviakov: `log_severity` is not captured on initialization and having two options for debug is a bit confusing. I have now changed `debug_logs` to boolean or string so that on `True` we just do max level logging and on `"INFO"`/`"DEBUG1"`/...  we do `log-severity`-like logging. Would it be a good solution?

kurapov-peter (approved):
LGTM

python/pyhdk/_common.pyx
```diff
@@ -199,14 +199,17 @@ def buildConfig(*, log_dir="hdk_log", **kwargs):
...
   cdef char *cargv0 = argv0
   cdef string default_log_dir = log_dir
   cdef unique_ptr[CLogOptions] opts = make_unique[CLogOptions](cargv0, default_log_dir)
   cmd_str = "".join(' --%s %r' % arg for arg in kwargs.iteritems())
   cmd_str = cmd_str.replace("_", "-")
+  if debug_logs is not None:
+    if isinstance(debug_logs, str):
+      cmd_str += ("--log-severity="+debug_logs)
+    elif debug_logs:
+      opts.get().severity_ = CSeverity.DEBUG3
```
ienkovich, line 213: Did you check it's not overwritten with the default value in the following `parse_command_line` call?

ienkovich (commented):
Looks good! But please make sure it's possible to pass scan nodes and execution results everywhere instead of a table name.

